PR #8, opened 2018-08-08
Implemented optional method to define log file paths

FelixMetzner:
Implemented optional method to define log file paths for tasks which are processed on a batch system and included instructions on how to use this feature as FAQ entry.


nils-braun:
Perfect PR @FelixMetzner. Good documentation, good bug fix.

PR #9, opened 2018-08-09
Using absolute path of executable in job submission command

FelixMetzner:
As the title suggest, this PR changes the job submission command so that the absolute path of the executable is used instead of only its name.


nils-braun:
:+1:

PR #14, opened 2018-11-29
 Adding a usage example for the product_dict

FelixMetzner:


b2luigi/core/utils.py
```diff
@@ -32,6 +32,12 @@ def product_dict(**kwargs):
         >>> list(product_dict(arg_1=[1, 2], arg_2=[3, 4]))
         [{'arg_1': 1, 'arg_2': 3}, {'arg_1': 1, 'arg_2': 4}, {'arg_1': 2, 'arg_2': 3}, {'arg_1': 2, 'arg_2': 4}]
 
+    The thus produced list can directly be used as inputs for a required tasks:
+
+    .. code-block:: python
+        def requires(self):
+            for args in list(product_dict(arg_1=[1, 2], arg_2=[3, 4])):
```
nils-braun: why the call to `list`?

nils-braun: Ah, you have probably copied the example from above. The `list` call was just to make it visible on the python console.

PR #17, opened 2019-07-03
Fix for `get_basf2_git_hash` to work with new tools

elimik31:
With the new BelleII tools, it seems that `$BELLE2_RELEASE` is only set if actually a release is used and is left unset for a development setup, where it head been set to `head` in the past. As a result, I got `None` as a git hash when running this function. I suppose it's no bug in the tools, so one could adapt this function here.

### DIFF
--- a/b2luigi/basf2_helper/utils.py
+++ b/b2luigi/basf2_helper/utils.py
@@ -6,7 +6,7 @@ import git
 def get_basf2_git_hash():
     basf2_release = os.getenv("BELLE2_RELEASE")
 
-    if basf2_release == "head":
+    if basf2_release == "head" or basf2_release is None:
         basf2_release_location = os.getenv("BELLE2_LOCAL_DIR")
 
         assert basf2_release_location


b2luigi/basf2_helper/utils.py
```diff
@@ -6,10 +6,10 @@
 def get_basf2_git_hash():
     basf2_release = os.getenv("BELLE2_RELEASE")
 
-    if basf2_release == "head":
+    if basf2_release == "head" or basf2_release is None:
```
elimik31: A shorter alternative would be 
```python
if basf2_release in ["head", None]:
```

b2luigi/basf2_helper/utils.py
```diff
@@ -6,10 +6,10 @@
...
         basf2_release_location = os.getenv("BELLE2_LOCAL_DIR")
 
         assert basf2_release_location
```
elimik31: This assertion will fail if basf2 is not set up and this function is called, as the release will also be `None` in that case. This is a change in behavior, as prior to that the function previously `None` when basf2 is not set up, but I don't think it is a problem.

nils-braun:
Thanks for the PR!

PR #18, opened 2019-07-09
Fix serialized parameters for basf2 tasks

elimik31:
As it is now, b2luigi.DispatchableTask has no method
`get_serialized_parameters`, therefore calling
```python3
super().get_serialized_parameters()
```
in `Basf2Task` does not work. However, there is a utility function with the same
name, which I just inserted here

Alternatively, one could add the utility function as a method to the
`DispatchableTask`, or remove the `get_serialized_parameters` method from `Basf2Task`
altogether to force people to use the utility function.


nils-braun:
Yes, you are correct. At some point I changed the API to the free function form but it seems I have missed this. Thanks for fixing.

PR #19, opened 2019-10-18
Feature/htcondor

welschma:
Hey,
I tested this the HTCondor feature on the NAF and our local cluster in Bonn. Until now I am not able to provide a working example script because its a bit complicated to setup the environment (I have to install the new b2luigi version again on the job execution node. I can provide the scripts and settings I used on the NAF and we could discuss how to add an example or improve the documentation. I also wrote Cyrill Praz if he could test it, but haven't heard from him yet.

Cheers Max


b2luigi/core/utils.py
```diff
@@ -240,6 +249,8 @@ def get_log_file_dir(task):
     base_log_file_dir = get_setting("log_folder", default=os.path.join(os.path.dirname(filename), "logs"))
 
     log_file_dir = create_output_file_name(task, task.get_task_family() + "/", create_folder=True, result_path=base_log_file_dir)
+    if not os.path.isdir(log_file_dir):
+        os.makedirs(log_file_dir, exist_ok=True)
```
nils-braun: As you check for the `isdir` above, I would argue the `exist_ok` should not be set in this case.

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
     your scheduling job.
 *   You have to choose the queue depending in your requirements (e.g. wall clock time) by yourself. So you need to make
     sure that the tasks will actually finish before the batch system kills them because of timeout.
-*   There is currently now resubmission implemented. This means dying jobs because of batch system failures are just
+*   There is currently no resubmission implemented. This means dying jobs because of batch system failures are just
```
nils-braun: Thanks!

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
...
+	"result_path": "/path/to/results",
+	"log_folder":  "/path/to/logs",
+	"env_setup": "/path/to/env_setup.sh",
+	"htcondor_settings": {
+            "+ContainerOS": "Ubuntu1804"
+	    }
+
+    }
+
+Additionaly, you have to provide a ``env_setup`` script that setups the working environment on the execution node
```
nils-braun: setups -> sets up

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
...
+(in contrast to LSF, which uses the same environment as the submission node). It will be automatically sourced before
+each ``luigi`` task is executed.
+Another new setting is ``htcondor_settings``. This is a dictionary where you can specify settings that put in the
```
nils-braun: will be put

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
...
+
+If you have job (task) specific settings, you can specify them by adding a ``htcondor_settings`` attribute
+to your task like this
+
+.. literalinclude:: ../../tests/htcondor/htcondor_example_b2luigi_2.py
+       :linenos:
+
+By assigning a dictionary to the ``htcondor_settings`` attribute (line 8-11 and 25-28), you can specify e.g the number of cpus or the
+required memory for the task. These settings have to valid HTCondor settings you would normally write into a job
+submission file.
```
nils-braun: Super nice.
Maybe you can add a link to the HTCondor documentation?

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
...
+
+Currently, besides the actual task output, the ``results`` directory contains the HTCondor job submission file, the
+exectuable wrapper and a copy of the settings file.
```
nils-braun: executable

docs/usage/quickstart.rst
```diff
@@ -191,8 +189,53 @@ Although the batch mode has many benefits, it would be unfair to not mention its
...
+to your task like this
+
+.. literalinclude:: ../../tests/htcondor/htcondor_example_b2luigi_2.py
+       :linenos:
+
+By assigning a dictionary to the ``htcondor_settings`` attribute (line 8-11 and 25-28), you can specify e.g the number of cpus or the
+required memory for the task. These settings have to valid HTCondor settings you would normally write into a job
+submission file.
+
+Currently, besides the actual task output, the ``results`` directory contains the HTCondor job submission file, the
```
nils-braun: contains -> will also contain

tests/doc_examples/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-	"result_path": "results"
+	"result_path": "results",
```
nils-braun: Unfortunately, that is not valid JSON anymore

tests/htcondor/htcondor_example_b2luigi_2.py
```diff
@@ -0,0 +1,53 @@
+import b2luigi
+import random
+
+
+class MyNumberTask(b2luigi.Task):
+    some_parameter = b2luigi.Parameter()
```
nils-braun: You can also use an `IntParameter` here ;-)

tests/htcondor/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+	"batch_system": "htcondor",
+	"result_path": "/cephfs/user/mwelsch/results",
+	"log_folder":  "/cephfs/user/mwelsch/logs",
+	"env_setup": "/cephfs/user/mwelsch/bla.sh",
+	"htcondor_settings": {
+      "+ContainerOS": "Ubuntu1804"
```
nils-braun: Can you indent once here?

tests/htcondor/simple_example_b2luigi.py
```diff
@@ -0,0 +1,20 @@
+import b2luigi
```
nils-braun: Do we need this file?

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+                history_cmd.append(str(job_id))
+                output = subprocess.check_output(history_cmd)
+                output = output.decode()
+                output = output.lstrip("[\n").rstrip("\n]\n").split("\n,\n")
+                if output:
+                    for status_dict in output:
+                        dict_list.append(json.loads(status_dict))
+
+        for status_dict in dict_list:
+            self[status_dict["ClusterId"]] = status_dict["JobStatus"]  # int -> int
```
nils-braun: I do not understand this comment

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+from b2luigi.batch.processes import BatchProcess, JobStatus, BatchJobStatusCache
+from b2luigi.core.utils import get_log_file_dir
+
+
+class HTCondorJobStatusCache(BatchJobStatusCache):
+
+    def __init__(self):
+        super(HTCondorJobStatusCache, self).__init__()
+
+    def _ask_for_job_status(self, job_id: int = None):
```
nils-braun: Maybe it is worth having a small comment pointing out what you are doing here

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class HTCondorProcess(BatchProcess):
+    """
+    Reference implementation of the batch process for a HTCondor batch system.
+
+    We assume that the batch system shares a file system with the submission node you
+    are currently working on (or at least the current folder and the result/log directory
+    are also available there with the same path).
+    The environment has to be set up on its own new for each job. For that, a environment
```
nils-braun: remove "new"

nils-braun: an environment

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+    setup script has to be provided in the ``settings.json`` file.
+    If no own python cmd is specified, the task is executed with the current ``python3``
+    available after the environment is setup.
+    General settings (may be dependend on your HTCondor setup) that affect all jobs (tasks)
+    can be specified in the ``settings.json`` by adding a ``htcondor_settings`` entry.
+    Job specific settings, e.g. number of cpus or required memory can be specified by adding
+    a ``htcondor_settings`` attribute to the task. It's value has to be a dictionary containig
```
nils-braun: containing

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+
+class HTCondorProcess(BatchProcess):
+    """
+    Reference implementation of the batch process for a HTCondor batch system.
+
+    We assume that the batch system shares a file system with the submission node you
+    are currently working on (or at least the current folder and the result/log directory
+    are also available there with the same path).
+    The environment has to be set up on its own new for each job. For that, a environment
+    setup script has to be provided in the ``settings.json`` file.
```
nils-braun: Can I also have this as a task-specific setting?
Can I also just give you a list of env variables (as in the LSF task case)?

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+        elif job_status == HTCondorJobStatus.idle:
+            return JobStatus.running
+        elif job_status == HTCondorJobStatus.running:
+            return JobStatus.running
+        elif job_status == HTCondorJobStatus.removed:
+            return JobStatus.aborted
+        elif job_status == HTCondorJobStatus.held:
+            return JobStatus.aborted
+        elif job_status == HTCondorJobStatus.submission_err:
+            return JobStatus.aborted
```
nils-braun: Maybe you could refactor this into
```
if job_status in [HTCondorJobStatus.completed]:
     return JobStatus.successful
elif job_status in [HTCondorJobStatus.idle, HTCondorJobStatus.running]:
     return JobStatus.running
elif job_status in [HTCondorJobStatus.removed, HTCondorJobStatus.held, HTCondorJobStatus.submission_err]:
    return JobStatus.aborted
```

Question (because I do not know it better): is it worth to have an Exception thrown on `submission_err`?

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+                output = output.lstrip("[\n").rstrip("\n]\n").split("\n,\n")
+                if output:
+                    for status_dict in output:
+                        dict_list.append(json.loads(status_dict))
+
+        for status_dict in dict_list:
+            self[status_dict["ClusterId"]] = status_dict["JobStatus"]  # int -> int
+
+
+class HTCondorJobStatus(enum.IntEnum):
```
nils-braun: Are those taken from the official HTCondor docu? Maybe have a link here (in case they need to be updated)?

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+            raise ValueError(f"Unknown HTCondor Job status: {job_status}")
+
+    def start_job(self):
+        submit_file_path = self._create_htcondor_submit_file()
+        exec_cmd, curr_env = self._create_condor_executable_cmd()
+
+        submit_file_dir, submit_file = os.path.split(submit_file_path)
+        cmd = ["condor_submit", submit_file]
+
+        # apparently you have to in the same directory as the submit file
```
nils-braun: to be in

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+        # to be able to submit jobs with htcondor
+        cur_dir = os.getcwd()
+
+        # have to copy settings file to job working directory since b2luigi has to
+        # take the result path from it
+        shutil.copyfile("settings.json", os.path.join(submit_file_dir, "settings.json"))
+        os.chdir(submit_file_dir)
+        output = subprocess.check_output(cmd, env=curr_env)
```
nils-braun: This command also has a cwd parameter, so you do not need to change the pwd on your own

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+
+        general_settings = get_setting("htcondor_settings", dict())
+
+        try:
+            job_settings = self.task.htcondor_settings
+        except AttributeError:
+            job_settings = {}
+
+        with open(submit_file_path, "w") as submit_file:
+            submit_file.write(f"executable = {executable_wrapper_path}\n")
```
nils-braun: Maybe worth having a multiline string here?
```python
submit_file.write(f"""
executable = ...
...
    """)

File: b2luigi/batch/processes/htcondor.py [hunk 22]
```diff
@@ -0,0 +1,225 @@
...
+        output_path = get_output_dirs(self.task, create_folder=True)
+        shell = get_setting("shell", "bash")
+
+        executable_wrapper_path = os.path.join(output_path, "executable_wrapper.sh")
+
+        condor_executable_cmd, _ = self._create_condor_executable_cmd()
+        condor_executable_cmd = " ".join(condor_executable_cmd)
+        with open(executable_wrapper_path, "w") as exec_wrapper:
+            exec_wrapper.write(f"#!/bin/{shell}\n")
+            exec_wrapper.write(f"source {env_setup_path}\n")
```
nils-braun: You could output the current env here if you want (but not needed, the user can also do this)

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+            exec_wrapper.write(f"{condor_executable_cmd}\n")
+
+        # make wrapper executable
+        st = os.stat(executable_wrapper_path)
+        os.chmod(executable_wrapper_path, st.st_mode | stat.S_IEXEC)
+
+        return executable_wrapper_path
+
+    def _create_condor_executable_cmd(self):
+
```
nils-braun: I think there is already a function like this, or? Maybe we can merge them?

b2luigi/batch/processes/htcondor.py
nils-braun: The `env` set by the task will then only be used during submission, correct? Is this worth mentioning in the docu? We could have a little chat on the env if you want

b2luigi/batch/processes/htcondor.py
welschma: It"s just a reminder that both, the key and the value in this dictionary are Integers. When i
extracted the `job_id` after the submission via a search using a regular expression, the `job_id` was given as string (see line 133). Checking the status in the cache didn't work back then, so i decided to cast the `job_id` to a `int`.

b2luigi/batch/processes/htcondor.py
welschma: The documentation states that we could also specify environment variables in the job description file. I also found a `getenv` option which should copy all current variables. I have to check again if in some cases one can ommit setting up a new environment.

b2luigi/batch/processes/htcondor.py
nils-braun: Nice

b2luigi/core/utils.py
welschma: removed the exist_ok argument

b2luigi/batch/processes/htcondor.py
welschma: Added the link and removed an outdated state :)

b2luigi/batch/processes/htcondor.py
welschma: In principle it's the same function. In the original function, the current python executable is used to execute the task on the worker node (in the LSF case). At the NAF I got some errors with the python libraries (I could try to reproduce them, but I assume that  the OS can differ from the submission and the worker node, at the NAF both SL6 and CentOS7 are used on the worker nodes if the OS is not specified). So this function is a workaround by using the current python3 command after setting up the environment. I suppose that dealing with the job environment is the biggest problem at the moment. I could read up a little and rethink the approach of providing the correct one.

b2luigi/batch/processes/htcondor.py
welschma: Added the multiline string form. Is there any other advantage of doing it this way than that it"s easier to read?

b2luigi/batch/processes/htcondor.py
welschma: Is refactored

b2luigi/batch/processes/htcondor.py
nils-braun: I guess your approach is fine, so we can keep it. I was more concerned about having two functions with more or less the  same code. So maybe it is possible to have an argument to the original function, so it can be reused as much as possible.

b2luigi/batch/processes/htcondor.py
nils-braun: No, really just readability. Thanks!

b2luigi/batch/processes/htcondor.py
welschma: I added an default argument to the `create_cmd_from_task` function to avoid code duplication. This function has been removed in commit f96812e

nils-braun: Nice!

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,225 @@
...
+        executable_wrapper_path = os.path.join(output_path, "executable_wrapper.sh")
+
+        condor_executable_cmd, _ = self._create_condor_executable_cmd()
+        condor_executable_cmd = " ".join(condor_executable_cmd)
+        with open(executable_wrapper_path, "w") as exec_wrapper:
+            exec_wrapper.write(f"#!/bin/{shell}\n")
+            exec_wrapper.write(f"source {env_setup_path}\n")
+            exec_wrapper.write("echo 'Starting executable'\n")
+            exec_wrapper.write("cd /jwd\n")
```
nils-braun: Sorry for bringing this up very late. What is `/jwd`?

nils-braun:
So (except for the `/jwd`) I am now in favor of merging this. We can still iterate later on the code itself for some generalisation and refactoring, but this feature is so nice I want to have it better now than later :-)

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,226 @@
...
+            exec_wrapper.write(f"{condor_executable_cmd}\n")
+
+        # make wrapper executable
+        st = os.stat(executable_wrapper_path)
+        os.chmod(executable_wrapper_path, st.st_mode | stat.S_IEXEC)
+
+        return executable_wrapper_path
```
nils-braun: I think you need to give the `abspath` of that, otherwise the default of just using "results" as the folder does not work.

b2luigi/batch/processes/htcondor.py
```diff
@@ -0,0 +1,226 @@
...
+        submit_file_dir, submit_file = os.path.split(submit_file_path)
+        cmd = ["condor_submit", submit_file]
+
+        # apparently you have to be in the same directory as the submit file
+        # to be able to submit jobs with htcondor
+        cur_dir = os.getcwd()
+
+        # have to copy settings file to job working directory since b2luigi has to
+        # take the result path from it
+        shutil.copyfile("settings.json", os.path.join(submit_file_dir, "settings.json"))
```
nils-braun: Just a small notice: that will not work for all cases. We can leave it as it is, I will try to come up with another solution.

b2luigi/batch/processes/htcondor.py
welschma: I guess it's good that you noticed that. It was supposed to be the job's working directory. Seems that this is special to the Bonn cluster. On the NAF this doesn't exist since it uses the submission directory as initial directory when the job starts running.

nils-braun: I think you can use a
```
initialdir = {os.getcwd()}
```
in the submission file to work against this. I am currently testing on the NAF and our local system. 
I would go with these changes:
```
diff --git a/b2luigi/batch/processes/htcondor.py b/b2luigi/batch/processes/htcondor.py
index 0126e23..24d4ba1 100755
--- a/b2luigi/batch/processes/htcondor.py
+++ b/b2luigi/batch/processes/htcondor.py
@@ -140,13 +140,6 @@ class HTCondorProcess(BatchProcess):
         submit_file_dir, submit_file = os.path.split(submit_file_path)
         cmd = ["condor_submit", submit_file]
 
-        # apparently you have to be in the same directory as the submit file
-        # to be able to submit jobs with htcondor
-        cur_dir = os.getcwd()
-
-        # have to copy settings file to job working directory since b2luigi has to
-        # take the result path from it
-        shutil.copyfile("settings.json", os.path.join(submit_file_dir, "settings.json"))
         output = subprocess.check_output(cmd, env=self.task_env, cwd=submit_file_dir)
         output = output.decode()
         match = re.search(r"[0-9]+\.", output)
@@ -180,11 +173,11 @@ class HTCondorProcess(BatchProcess):
         with open(submit_file_path, "w") as submit_file:
             submit_file.write(f"""
             executable = {executable_wrapper_path}
-            should_transfer_files = YES
-            transfer_input_files = settings.json
+            initialdir = {os.getcwd()}
             log = {job_log_file}
             output = {stdout_log_file}
             error = {stderr_log_file}
+            transfer_output_files = ""
             """)
 
             for key, item in general_settings.items():
@@ -199,7 +192,7 @@ class HTCondorProcess(BatchProcess):
 
     def _create_executable_wrapper(self):
 
-        env_setup_path = get_setting("env_setup")
+        env_setup_path = os.path.abspath(get_setting("env_setup"))
 
         if not os.path.isfile(env_setup_path):
             raise FileNotFoundError(f"Environment setup script {env_setup_path} does not exist.")
@@ -221,6 +214,6 @@ class HTCondorProcess(BatchProcess):
         st = os.stat(executable_wrapper_path)
         os.chmod(executable_wrapper_path, st.st_mode | stat.S_IEXEC)
 
-        return executable_wrapper_path
+        return os.path.abspath(executable_wrapper_path)
```
That should fix all problems with files and folders...

nils-braun: If you use the current folder as start folder, you also do not need to copy the settings file :-)

nils-braun: Narf. of course the initaldir does not work...
Anyways, two possibilities: we could just "cd" into the current folder or we can create out own settings file. Will think about it

nils-braun:
I finally start to understand how stupid this all is :-D Makes your PR even more valuable.
I think we merge it now and reconvene later (maybe also on Friday). I will do another PR with some suggestions after that.

PR #23, opened 2019-11-14
corrected path to decfile fore new structure in release-04

welschma:
The location of the ccbar decay file change in release 04. So the basf2 example won't work at the moment


nils-braun:
Thanks!

PR #24, opened 2020-01-10
include task name in directory structure of the created output file name

welschma:
If applied, this commit will include task name in directory structure of the created output file name.

At the moment, the structure of the results directory only depends on the task parameters but not the task (names) itself. This change will add the task name before the parameter list.

### DIFF
--- a/b2luigi/core/utils.py
+++ b/b2luigi/core/utils.py
@@ -229,7 +229,7 @@ def create_output_file_name(task, base_filename, result_dir=None):
         result_dir = map_folder(get_setting("result_dir", task=task, default=".", deprecated_keys=["result_path"]))
 
     param_list = [f"{key}={value}" for key, value in serialized_parameters.items()]
-    output_path = os.path.join(result_dir, *param_list)
+    output_path = os.path.join(result_dir, task.get_task_family(), *param_list)
 
     return os.path.join(output_path, base_filename)
 


welschma:
if you think this should be the case, i would then change the failing tests accordingly

meliache:
So as I understand it, currently (before this PR) when multiple tasks have the same output file name with the same luigi parameters, the task output is marked _DONE_ for all tasks if at least one of the tasks created that output. But it's hard to think of a case where that might be a desired behavior.

With your PR the task name is part of the file structure and thus you ensure that the output is defined uniquely for each task, even if the output names are identical and parameters for different tasks are identical. Did I understand that correctly?

As I understand, this commit also will change the way how people navigate the directory structure. At the moment I  can go into some subdirectory defined by the training parameters, e.g. `n_events_training=10000/n_events_testing=1000` and see the outputs of multiple tasks. This commit willl change the behavior that I first have to go into the task subdirectory and there descend the parameter directories.

I don't care too much, don't think my old b2luigi scripts would be influenced in any negative way.

welschma:
At the moment i wanted to compare some analysis results for to cases, Generator Level and Reconstruction Level on NTuple level. For convenience, I created two different Tasks with the same parameters that create the plots for the comparison. The tasks take the same Ntuples as input and do mostly the same things but it was easier to define all variable, label, title names easily by hand for the different plots. 
Maybe the structure of my tasks is just the problem here or i should just not repeat the file output file names in different tasks

nils-braun:
So, First of all thank you for the PR.
I am basically with both of you: I also see your point and I see where this could be useful.

However, I am a bit reluctant to include this change and the reason is the one described by Michael: I intentionally wanted the feature that all results with the same parameters go into the same folder because they "mean" the same thing.
 The way I solved your use-case is to add another parameter (e.g. named tag or label) with a fixed value (one could even think about a constant parameter class, which is immutable). But I see that this is also not optimal.
Anyways, I am open for this idea and would give back the decision to the community.
In all cases, a b2luigi setting to control that sounds good (we only need to decide on the default)

nils-braun:
But one suggestion unrelated to the decision: could you move the task class name to the end? I think that makes a bit more sense in the structure and it is in line in how the submission files are created

welschma:
I will revisit my structure of the task regarding the final analysis steps on monday. Probably this would easily solve my issue.

PR #25, opened 2020-01-12
Adding option for user-defined get_task_file_dir

FelixMetzner:
This PR introduces the option to also provide a user-defined directory for the task executable via `get_task_file_dir` analog to the option provided for `get_log_file_dir`.

If a task defines the method `get_task_file_dir` the return value of this method will be interpreted as path to the directory to be used for the location of the executable file.

I also thought about using one optional task method such as `get_task_working_dir` to be used in both, the `get_log_file_dir` and the `get_task_file_dir`, but refrained from doing so, because this directory does not necessarily have to be the working directory. In my use case this directory only contains information for and about the task execution on the LSF batch system. For the htcondor case the directory fulfills a  more important purpose.

If you have suggestions for improvements, please let me know.

### DIFF
--- a/b2luigi/core/utils.py
+++ b/b2luigi/core/utils.py
@@ -9,7 +9,8 @@ import types
 
 import colorama
 
-from b2luigi.core.settings import set_setting, get_setting
+from b2luigi.core.settings import get_setting
+
 
 @contextlib.contextmanager
 def remember_cwd():
@@ -238,9 +239,9 @@ def get_log_file_dir(task):
     if hasattr(task, 'get_log_file_dir'):
         log_file_dir = task.get_log_file_dir()
         return log_file_dir
-        
+
     # Be sure to evaluate things relative to the current executed file, not to where we are now
-    base_log_file_dir = map_folder(get_setting("log_dir", task=task, default="logs", 
+    base_log_file_dir = map_folder(get_setting("log_dir", task=task, default="logs",
                                                deprecated_keys=["log_folder"]))
     log_file_dir = create_output_file_name(task, task.get_task_family() + "/", result_dir=base_log_file_dir)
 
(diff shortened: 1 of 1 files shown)


b2luigi/core/utils.py
```diff
@@ -9,7 +9,8 @@
 
 import colorama
 
-from b2luigi.core.settings import set_setting, get_setting
+from b2luigi.core.settings import get_setting
```
nils-braun: Thanks for fixing this!

b2luigi/core/utils.py
```diff
@@ -238,16 +239,20 @@ def get_log_file_dir(task):
...
-    base_log_file_dir = map_folder(get_setting("log_dir", task=task, default="logs", 
+    base_log_file_dir = map_folder(get_setting("log_dir", task=task, default="logs",
                                                deprecated_keys=["log_folder"]))
     log_file_dir = create_output_file_name(task, task.get_task_family() + "/", result_dir=base_log_file_dir)
 
     return log_file_dir
 
 
 def get_task_file_dir(task):
+    if hasattr(task, 'get_task_file_dir'):
```
nils-braun: Yes, I am fine with that!

I thought about also having a line like `base_dir = ...` as in the `get_log_file_dir` function, but that would duplicate code.
So lets go with that!

nils-braun:
Looking at it (not your thing, more my fault), the naming is now a bit inconsistent.
If you want to change the log folder, it is `get_log_file_dir` (good), or it uses the base folder `log_dir` (good).
If you want to change the task folder, it is `get_task_file_dir` (good), or it uses the base folder `result_dir` (not good, inconsistent).
Anyways, that's how it is :man_shrugging:

PR #26, opened 2020-02-19
When creating output file names, escape forward slashes by dots

meliache:
When a luigi parameter contains slashes, e.g. when it is a path, additional
subdirectories for each substring between the slashes in the parameter are created. It actually does not break anything, because the getting and creating of filenames is still symmetric. I am not sure what the best way to escape the path is, I chose dots but can change that if you prefer something else.

Found that when I used b2luigi to send gbasf2 jobs and passed grid datasets as a parameter and was surprised by the output directory structure.

### DIFF
--- a/b2luigi/core/utils.py
+++ b/b2luigi/core/utils.py
@@ -230,7 +230,8 @@ def create_output_file_name(task, base_filename, result_dir=None):
         result_dir = map_folder(get_setting("result_dir", task=task, default=".", deprecated_keys=["result_path"]))
 
     param_list = [f"{key}={value}" for key, value in serialized_parameters.items()]
-    output_path = os.path.join(result_dir, *param_list)
+    excaped_param_list = [param.replace("/", ".") for param in param_list]
+    output_path = os.path.join(result_dir, *excaped_param_list)
 
     return os.path.join(output_path, base_filename)
 


b2luigi/core/utils.py
```diff
@@ -230,7 +230,8 @@ def create_output_file_name(task, base_filename, result_dir=None):
         result_dir = map_folder(get_setting("result_dir", task=task, default=".", deprecated_keys=["result_path"]))
 
     param_list = [f"{key}={value}" for key, value in serialized_parameters.items()]
-    output_path = os.path.join(result_dir, *param_list)
+    excaped_param_list = [param.replace("/", ".") for param in param_list]
```
meliache: That `replace` could also be done in-place inside the first list comprehension in the line above, but for readability I thought it's better to split it up and have a separate statement on a separate line.

FelixMetzner:
You can also use the `hashed` option for the respective parameter as described [here in the FAQs](https://b2luigi.readthedocs.io/en/latest/advanced/faq.html#how-do-i-handle-parameter-values-which-include-or-other-unusual-characters). certain parameters are just not suitable as directory names...

nils-braun:
Thanks for the PR! 
I am not a huge fan of "just doing this" without giving the user the opportunity to choose. Two reasons for this:
* introducing the change will break every output file structure that already contains a "/" and people would need to redo their calculations
* when reading back the file path and extracting the parameters you would end up with another parameter value then what was used during writing

How about the following ideas:
* we create another type of parameter, say `EscapedParameter` which does the replacement automatically (all parameters in luigi have a method which is called to transform from bare parameter value to stringified parameter value and one could hook that). Users can then choose this parameter on purpose.
* we include another option in the parameter (same what was done for `hashed`)
* or (as @FelixMetzner suggested) you just use `hashed` which was introduced exactly because of this.

b2luigi/core/utils.py
nils-braun: There is a typo: "excaped" -> "escaped" (I guess this is what you wanted to say)

meliache: Yes, meant that.

meliache:
Thanks for the suggestions and sorry for missing the FAQ, should have thought about hashed. But since strings are valid Parameter values in luigi, I assumed they should work without braking in b2luigi. 

There is the "always sanitize user input" mantra, but I doubt it applies in this case (it's a dev tool and not an online form). But if we were to introduced escaped parameters or a corresponding option, one should escape all characters that are not allowed in linux path names, so the simple slash-escape proposed here is not enough if you want to do it properly.

As you said, if you need them, creating your own parameters is fairly easy in luigi.

PR #27, opened 2020-02-28
Feature/warning if forward slash in parameter

meliache:
Print a warning suggesting to use hashed parameters when a luigi parameter value has a forward slash, as this will split the output path in multiple directories corresponding to the substrings on both sides of the slash. Not an Exception, as b2luigi will still work.

The warning is issued in `create_output_file_name` so that people only see it if they use the `get_output_file_name` utilities of b2luigi.

Related to https://github.com/nils-braun/b2luigi/pull/26

### DIFF
--- a/b2luigi/core/utils.py
+++ b/b2luigi/core/utils.py
@@ -6,6 +6,7 @@ import os
 import collections
 import sys
 import types
+import warnings
 
 import colorama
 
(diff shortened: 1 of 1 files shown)


b2luigi/core/utils.py
```diff
@@ -229,6 +230,12 @@ def create_output_file_name(task, base_filename, result_dir=None):
         # Be sure to evaluate things relative to the current executed file, not to where we are now
         result_dir = map_folder(get_setting("result_dir", task=task, default=".", deprecated_keys=["result_path"]))
 
+    for key, value in serialized_parameters.items():
```
meliache: Now that I included a for loop over the parameters, I could also build the `param_list` in the same loop instead of the separate list comprehension below.

b2luigi/core/utils.py
```diff
@@ -229,6 +230,12 @@ def create_output_file_name(task, base_filename, result_dir=None):
...
+        if isinstance(key, str) and "/" in value:
+            warnings.warn(f"Value of parameter ``{key}`` contains forward slash \"/\". "
```
meliache: This is a generic warning, if you want I could define a new warning class, similar as you did for the `DeprecationSettingsWarning`

nils-braun:
Nice, thanks for the PR. I am fine with the changes as they are right now. If you are ok, I will merge

meliache:
I am fine with the warning as it is now, if nobody else has issues with that, I am fine with you merging.

PR #30, opened 2020-04-16
Feature/additional requirements structure

welschma:
This pull request implements a new function that allows the handling of dictionary returns of the requirement method. 
This is useful in situations where you require the same task with different parameters, where you also want to handle the input data differently depending on the parameter (e.g. if you preprocess some MC and at some point you want to differentiate signal and background samples).


nils-braun:
Sorry, I really just forgot about this PR :-/
It is fine!

PR #32, opened 2020-04-24
Soft wrapper for gbasf2 as a b2luigi BatchProcess

meliache:
A batch process that wraps gbasf2 commands to submit jobs to the grid.
Build of the sphinx documentation is hosted [here](https://meliache.github.io/b2luigi/docs/_build/html/usage/batch.html#gbasf2-wrapper-for-lcg).

It only works for tasks that have a `create_path()` method, usually `Basf2PathTask`s. It therefore has basf2 as a dependency and uses the path serialization features included in basf2 to store the path created by this method on disk in the `task_file_dir`. It uses a jinja template to create a minimal wrapper steering file that loads that path and processes it. Therefore, as of now, it only supports pickable analyses. Then, it sends both the path pickle and the steering file wrapper to the grid via by calling `gbasf2` in a python subprocess.

gbasf2 commands don't work in the normal basf2 environment, but usually the user has to source a gbasf2 setup script which is incompatible with basf2 and execute the gbasf2 commands (which are python2) in that environment. The way that issue is solved is that the BatchProcess has a `gbasf2_env` property function. It sources the gbasf2 setupt script in a subprocess and store the output of the bash `env` command in a string, from which a python dictionary with the environment is created. Then, I execute all my gbasf2 subprocesses via
```python
subprocess.run(<cmd>, env=gbasf_env, ...)
```

Mostly I work with the output strings from the gbasf2 commands. This is a bit dangerous, as outputs might change and furthermore, in most cases, gbasf2 commands don't raise and exception or return an error code when something unexpected happens, but print an error message via stdout. For analysing the status of a gbasf2 project, I actually wrote a python2 file which communicates with the BelleDirac API and returns a json with statuses of all jobs in the project. This python2 file is also then run in a subprocess with the gbasf2 env.

This gbasf2 batch process checks regularly what the status of the jobs on the grid is and prints a summary with the numbers of jobs in each job state. I made it so that this prints only when the numbers change and it can be disabled via the `gbasf2_print_status_updates` setting. I am very open to suggestion on how to improve my monitoring output or logging. When jobs fail, it uses the `gb2_job_reschedule` command to reschedule them and stores the number of tries per job in a json file. If the number of tries of a job is larger equal the `gbasf2_max_retries` setting (defaults to 0), it doesn't reschedule the job and `get_job_status` is failed. A successfull gbasf2 project requires all jobs in it to be successful.

For gbasf2 projects, the user normally has to provide a project name (alphanumeric, <= 32 chars). For that to be unique, I let the user provide a prefix via a setting and append to the name of the prefix a hash of the task id.

I make a lot of use of settings, all are top-level b2luigi settings and start with `gbasf2_...`. I don't think that's very elegant, I would prefer to have in the settings file something like
```
{"gbasf2": {"input_dataset": ...}}
```
But I want tio be able to retrieve the settings via `get_setting` directly and for almost all settings I want to be able to provide default values. Further, it's useful to be able to set settings as task class properities or luigi parameters.

One other thing I don't like is that usually the output of a task shouldn't depend on the batch system. However, gbasf2 automatically creates multiple jobs for a task and then downloads the output as a directory. And it is not really determenistic, how the file in that directly are colled. Usually, when the basf2 path outputs an `ntuple.root`, it will result in gbasf2 output files named `ntuple_1.root`, `ntuple_2.root`, `ntuple_n.root`, where `n` is the number of files in the input dataset (that I can obtain). However, I don't know if things are always like that when jobs get rescheduled or for example if the user manually deletes certain jobs because he wants the project to be successful even though those tasks failed. So in my example, I currently define the output directory as the output of the task, which seems the most straighforward simple solution. To avoid problems when the download fails (frequent), I download the datasets and the logs first into temporary directories and then move them on success to their final location.
 
Okay now I'm tired of writing the PR, I think you can look at the code and will already find something to comment on. I will try to add some example outputs later on so that you can get an idea of how it looks to run it. Max is currently trying it.


pyproject.toml
```diff
@@ -9,4 +9,4 @@ author-email = "nils.braun@kit.edu"
 home-page = "https://github.com/nils-braun/b2luigi"
 classifiers = ["License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)"]
 
-requires=['luigi<2.7.9', 'parse>=1.8.4', 'GitPython>=2.1.11', "colorama>=0.3.9", "cachetools>=2.1.0"]
+requires=['luigi<2.7.9', 'parse>=1.8.4', 'GitPython>=2.1.11', "colorama>=0.3.9", "cachetools>=2.1.0", "jinja2"]
```
meliache, line 12: Maybe I should also add a version number to my jinja2 requirement? I use it in one place and its a really basic use-case, I would expect it to work with all versions, but I don't know.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,673 @@
...
+
+        #: Maximum number of times that each job in the project can be rescheduled until the project is declared as failed.
+        self.max_retries = get_setting("gbasf2_max_retries", default=0, task=self.task)
+
+        #: Store number of times each job had been rescheduled
+        self.n_retries_by_job = Counter()
+
+        #: Local storage for ``n_retries_by_job`` counter
+        # so that it persists even if luigi process is killed and restarted.
+        # TODO: Maybe a small database (e.g. tindydb/pickledb) would be more appropriate
```
meliache, line 199: This is the only TODO comment left in my code. I use a `collections.Counter()` for storing the number of retries for each jobid and whenever I change the counter, I write all of it to the json file. This overwriting of the whole file seems inefficient to me, as I just want to change one single value in that json. Not sure if I can do that easily with json. One idea would be a minimalistic DB, but that would add a dependency and maybe add code complexity, as I just want a key-value store.

welschma:
I've tested the example and its working fine for me. Haven't looked at the code yet :)

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,678 @@
...
+            return JobStatus.aborted
+
+        if n_in_final_state < n_jobs:
+            return JobStatus.running
+
+        # Require all jobs to be done for project success, any job failure results in a failed project
+        if n_done == n_jobs:
+            # download dataset only the first time that we return JobStatus.successful
+            if not self._project_had_been_successful:
+                self._on_first_success_action()
```
meliache, line 332: So the way I solved inititating the project dataset download from the grid after all jobs had been successful is to add the code for that as a side effect into the `get_job_status` method, which on success triggers the `on_first_success_action`. I don't really like that the `get_job_status` does more than just getting the job status. maybe there's a better way to do that, if you know some luigi functions I could implement I would be glad to do it there.

meliache:
The idea came to my mind that people might not want to download the output datasets for all their tasks. E.g. if you want to do some data/MC production on the grid and provide files to other people or if you want to chain gbasf2 tasks. Then, each gbasf2 tasks depends on the grid output of the previous, but doesn't required any local tasks.
- [ ] Add option to disable downloading the output dataset on project success
- [ ] Think how the user would then have to define his output function, maybe provide another target or override the `task.complete` method. Maybe I should provide helper functions for that.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+            steering_file_stream.dump(self.wrapper_file_path)
+
+    def _download_dataset(self):
+        """
+        Download the results from a gbasf2 project, stored as a dataset on the grid.
+        """
+        # Get list of files that we want to download from the grid via ``gb2_ds_list`` so that we can
+        # then compare this list with the results of the download to see if it was successful
+        ds_list_command = shlex.split(f"gb2_ds_list --user {self.dirac_user} {self.gbasf2_project_name}")
+        output_dataset_str = subprocess.run(ds_list_command, check=True, env=self.gbasf2_env, stdout=PIPE, encoding="utf-8").stdout
```
meliache, line 483: I used `subprocess.run(..., check=True, stdout=PIPE).stdout instead of `subprocess.check_output` in most cases, because the current subprocess documentation recommends to use the universal `run` high level interface. However, it's much more verbose in python 3.6, starting from 3.7 I could have just set the new `capture_output` argument of `run` to `True`.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+          the BelleII-specific Dirac-wrapper gbasf2.
+
+        - **Project status monitoring**
+
+          After the project submission, the gbasf batch process regularly checks the status
+          of all the jobs belonging to a gbasf2 project returns a success if
+          all jobs had been successful, while a single failed job results in a failed project.
+          You can close a running b2luigi process and then start your script again and if a
+          task with the same project name is running, this b2luigi gbasf2 wrapper will recognize that
+          and instead of resubmitting a new project, continue monitoring the running project.
```
nils-braun, line 43: Nice.
We should have this also for the other batch systems. Like that.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+        n_jobs_by_status = Counter()
+        for _, job_info in job_status_dict.items():
+            n_jobs_by_status[job_info["Status"]] += 1
+
+        # print summary of jobs in project if setting is set and job status changed
+        if (get_setting("gbasf2_print_status_updates", default=True, task=self.task) and
+                n_jobs_by_status != self._n_jobs_by_status):
+            time_string = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            job_status_string = str(dict(sorted(n_jobs_by_status.items()))).strip("{}")
+            print(f"Jobs in gbasf2 project \"{self.gbasf2_project_name}\" at {time_string}: {job_status_string}")
```
nils-braun, line 267: Would you think this is useful for all batch processes, e.g. also for htcondor? Then we could implement this in the calling classes.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+        # If project is does not exist on grid yet, so can't query for gbasf2 project status
+        if not check_project_exists(self.gbasf2_project_name, self.gbasf2_env, dirac_user=self.dirac_user):
+            raise RuntimeError(
+                f"Could not find any jobs for project {self.gbasf2_project_name} on the grid.\n" +
+                "Probably there was an error during the project submission when running the gbasf2 command.\n" +
+                "Try if you can run the gbasf2 command used manually in a terminal with gbasf2 set up:\n" +
+                " ".join(self._build_gbasf2_submit_command())
+            )
+
+        job_status_dict = get_gbasf2_project_job_status_dict(self.gbasf2_project_name, self.gbasf2_env, dirac_user=self.dirac_user)
```
nils-braun, line 257: As I do not know `gbasf2`: can you gain speed by asking for all projects in a single call and then caching it for the next seconds? If yes, you can use the `BatchJobStatusCache`.

b2luigi/batch/processes/gbasf2.py
nils-braun: Would it work with the luigi event handlers (https://luigi.readthedocs.io/en/stable/tasks.html#events)?
Although I am not sure we call them correctly in every circumstance (and if they are called on the submission side).
Just came to my mind without me thinking about it more clearly. Don't know if you had a look into it already.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+            template = Template(template_file.read())
+            # replace some variable values in the template
+            steering_file_stream = template.stream(
+                pickle_file_path=self.pickle_file_path,
+                max_event=get_setting("max_event", default=0, task=self.task),
+            )
+            # write the template with the replacements to a new file which should be sent to the grid
+            steering_file_stream.dump(self.wrapper_file_path)
+
+    def _download_dataset(self):
```
nils-braun, line 476: Maybe I do not get it, but why do you need an additional `gbasf2_output_directory`? Is the `result_dir` not enough - or put other way: why are you not using the locations of the file paths in the targets of `output()`?
Sorry, do not know gbasf2 enough,

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+        Create a steering file to send to the grid that executes the pickled
+        basf2 path from ``self.task.create_path()``.
+        """
+        # read a jinja2 template for the steerinfile that should execute the pickled path
+        template_file_path = os.path.join(self._file_dir, "templates/gbasf2_steering_file_wrapper.jinja2")
+        with open(template_file_path, "r") as template_file:
+            template = Template(template_file.read())
+            # replace some variable values in the template
+            steering_file_stream = template.stream(
+                pickle_file_path=self.pickle_file_path,
```
nils-braun, line 470: Is this path also available on the grid? I mean you copy the file on the submit call, but if I understand correctly you replace the template string with the full path here, or?

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+        """
+        try:
+            path = self.task.create_path()
+        except AttributeError as err:
+            raise Exception(
+                "Gbasf2 batch process can only used with tasks that generate basf2 paths with "
+                "a ``create_path()`` method, e.g. are an instance of ``Basf2PathTask``."
+            ) from err
+
+        path.add_module("Progress")
```
nils-braun, line 456: Please do not change the path of the user. I used to do this, but at some point users will have problems because the number of modules they expect is not the same.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,764 @@
...
+        Things to do after the project failed
+        """
+        job_status_dict = get_gbasf2_project_job_status_dict(self.gbasf2_project_name, self.gbasf2_env, dirac_user=self.dirac_user)
+        failed_job_dict = {job_id: job_info for job_id, job_info in job_status_dict.items()
+                           if job_info["Status"] == "Failed"}
+        n_failed = len(failed_job_dict)
+        print(f"{n_failed} failed jobs:\n{failed_job_dict}")
+        self._download_logs()
+
+    def _reschedule_failed_jobs(self):
```
nils-braun, line 317: I do like your idea. However, luigi already comes with a `retry` configuration (https://luigi.readthedocs.io/en/stable/configuration.html#per-task-retry-policy). Would this also work? I know, you implement a lot more functionality (e.g. the storage to files, which would only be possible by using a central scheduler in the luigi-only case) - but of course less own code is always better.

b2luigi/batch/processes/gbasf2.py
meliache: I use absolute paths, but gbasf2/dirac recreates the directory structure of the path in which gbasf2 is called on the grid inside the job input sandbox and probably chroot's into it or something like that.  So far it seemed to work for my use-cases. Relative paths would probably also have worked.

b2luigi/batch/processes/gbasf2.py
meliache: Thanks for spotting that, that's something where I was interested in your opinion but forgot to comment on that explicitly. In most cases that I had seen so far, the `output()` of a task doesn't influence the what happens in the `run()`, it is only used to define the targets of the task, but the target paths are acutally set e.g. in your `create_path()` in the `RootOutput` module or whatever. So I felt it's not supposed to be used for *setting* the output location. But the `create_path()` has also no influence where the downloaded gbasf2 dataset will finally go, so I decided to set it via a setting. It is set by default to the path you get from `get_output_file_name(...)`, but there might be users who don't want to use the output helpers and `result_dir` and for them I decided to provide this setting.

I feared ising `output()` to set the output directory path might also take away some flexibility from the user in using some other targets in the `output()` which are not the dataset directory. E.g. in a commit yesterday I added a `Gbasf2GridProjectTarget` which checks if the dataset exists on the grid and is finished, for jobs where the dataset download is disabled via a setting. But well, in that case particular case I would never execute the download code, so the fact the `output` is not a file system target wouldn't be a problem there, now that I think about it.

The output handling is something that I wasn't sure so I am happy about your suggestions. I thought maybe I could also add some helper functions for the user. But if you think my fears are no reason to worry about, I might just use `output()`.

gbasf2 is not complicated in that, thare is a `gb2_ds_get` command which takes a project name and by defaults downloads  the root files into `./<project name>/<output root files>`. You can also use a command line option (`-i`) to download it into another directory, but it will still have the project name in a subdirectory. I don't use that CLI option but just set the `cwd` argument in subprocess and then move/rename the directory to its final location.

examples/gbasf2/gbasf2_example.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Example luigi cut parameter to facilitate starting multiple projects for different cut values
+    mbc_lower_cut = b2luigi.IntParameter()
+
+    def create_path(self):
+        mbc_range = (self.mbc_lower_cut, 5.3)
+        return example_mdst_analysis.create_analysis_path(mbc_range=mbc_range)
+
+    def output(self):
+        yield self.add_to_output(self.gbasf2_project_name_prefix)
```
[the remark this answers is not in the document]
meliache: It's the default location of the directory to which I download the output dataset, which can be changed via the `gbasf2_output_directory` setting. I am not entirely happy with that, it's not really clear to the user. I discussed why I didn't set the output directory path in the `output()` in my long reply [below](https://github.com/nils-braun/b2luigi/pull/32#discussion_r416840833)

b2luigi/batch/processes/gbasf2.py
meliache: Didn't know about the event handlers I will look into that, thanks. I only knew about the `on_success` etc. methods of tasks and I didn't know how to hook into that.

b2luigi/batch/processes/gbasf2.py
meliache: Removed in ea01ba1

b2luigi/batch/processes/gbasf2.py
meliache: I retry single (sub-) jobs in a gbasf2 project and keep count of the number of retries for each individual jobs. The reason is that I regard a project a failed if a single job in it failed and then I want to only retry that job. If 100 jobs in a project fail, the global task/project retry counter would tell me that the task had 100 retries. Another possiblity would be to wait for all jobs to finish running, so they are either done or failed, and then retry all jobs in the project in one go (the `gb2_job_reschedule` command can send a retry command to all failed jobs in a project).

Actually, now that I think about it, having a cumulative per-job retry counter is not that bad, b/c if you have 100 failed jobs the user should look at his jobs, because then it's unlikely to be an issue with the grid site.

b2luigi/batch/processes/gbasf2.py
meliache: I added a `run_with_gbasf2` which runs `subprocess.run` with the gbasf2 environment and changed default settings. Also I added the `capture_output` argument from python 3.8.

b2luigi/batch/processes/gbasf2.py
nils-braun: I guess I do not understand all of this correctly (not due to the fact you explaining it wrongly, but due to the fact that I never touched any of that), but I have the feeling you are somehow "mis-using" luigi here.
The targets you define in your Task (basically what you return in `output()`) are there for two reasons in luigi:
1. define for any dependent task where to find its input files
2. tell luigi when a task is finished, basically iff all targets return true on their `exist()` call.

If I understand your comment correctly, the following could happend:
* a user defines some `LocalTarget` with a specific path in the `output()` call
* in the created path, the user specifies another output file name (which is already not how it should work)
* your code will download the created file into the folder specified by `gbasf2_output_directory`, or by default to `get_output_file_name(get_setting("gbasf2_project_name_prefix", task=self.task))`

If you now specify a `result_dir` of anything not being `gbasf2_project_name_prefix` (or `gbasf2_output_directory`), this task will never be finished for luigi, because its targets never exist.

Questions from my side:
* how does gbasf2 determine what it should download, e.g. what is in the "project dataset"?
* you said (in some command above) the full absolute file paths are regenerated inside of the grid environment. Could one just write files in the grid job already to their final location (as given by the path in the LocalTarget) and then download them and copy them exactly to where they belong?
* I did not quite understand your sentence "In most cases that I had seen so far, the output() of a task doesn't influence the what happens in the run()". Because looking into our examples (https://b2luigi.readthedocs.io/en/stable/advanced/basf2-examples.html) we tell the user to use the function "self.get_output_file_name" to specify the output file name.

If using "get_output_file_name" is a problem on grid jobs (due to whatever reason), I would argue to only allow for custom targets, such as your already defined target without download and to add one with download. It would be clearer to the user that these are not typical files but more a "flag" that the job is done. Still, you could give them a file `path` of the downloaded location (maybe controlled by your parameter) that would make them usable in any task depending this one.

Example:

```python
class MyTask(...):
    def output(self):
        yield Gbasf2GridProjectTargetWithDownload("ntuples.root") # this could check only locally 

    def create_path(self):
        ... RootOutput("ntuples.root")....

```

Still, I would definitely vote (if possible) for a solution like:

```python
class MyTask(...):
    def output(self):
        yield self.add_to_output("ntuples.root")
        # or, if the user really wants
        yield LocalTarget(.... ntuples.root ....)

    def create_path(self):
        ... RootOutput(self.get_output_file_name("ntuples.root"))....
   
   # or if this does not work

    def create_path(self):
        ... RootOutput("ntuples.root")....

```

The reason I strongly vote for this, is that this is what luigi is all about in its core: define targets and the definition of done for a task is, when someone has written exactly this target. Its much cleaner then.

So the way I would handle it: 
on downloading, loop through all targets in `task.output()`, check if they actually have a path, download the dataset into a temporary folder and move the files around.
You said above, that gbasf2 will create not a single output, but actually multiple ones (e.g. called `ntuple_X.root`). You could work around that by assuming the file paths given in the "output" are actually folder names. So in the example code above I would expect to have a folder called "ntuples.root" in my result_dir (with the usual parameters before that), which includes all the ntuples files.

What do you think about that? Does this fit your idea and how gbasf2 works? Sorry if I give stupid comments, please correct me.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -0,0 +1,782 @@
...
+from functools import lru_cache
+
+from luigi.target import Target
+from b2luigi.basf2_helper.utils import get_basf2_git_hash
+from b2luigi.batch.processes import BatchProcess, JobStatus
+from b2luigi.core.settings import get_setting
+from b2luigi.core.utils import get_log_file_dir, get_task_file_dir
+from jinja2 import Template
+
+import basf2.pickle_path as b2pp
```
nils-braun, line 20: Ah, sorry for the late comment: can you please make this a lazy import? Otherwise the core components of `b2luigi` can not be imported without having `basf2` - which also means all the tests fail.

b2luigi/batch/processes/gbasf2.py
meliache: Thanks for the detailed answer and suggestions!

1. **how does gbasf2 determine what it should download, e.g. what is in the "project dataset"?**

    On the grid, files can be accessed via a logical file name (LFN) which is a unix-like path. The output data of all user jobs is actually stored unter  `/belle/user/<user name>/<project>/file`. I just found out on confluence that there's also the possiblity to have subdirectories in special cases, I just copied a section:

    > `/belle/user/<user name>/<project>/file`
    > or
    > `/belle/user/<user name>/<project>/<subdir>/file`
    >
    > Subdirectories can be either
    > - the ones created by the option '--subdir_nfiles'
    > - the ones created following the DataDescription metadata in multi-output jobs
    >
    > File names (the basename) are combination of
    > - user-defined output file name
    > - job number

    The gbasf2 submit command also has cli option to store the outputs in a entirely different path, e.g when you want to provide data to your subgroup or do data production. But I provide only the project name in my submission command and expect the outputs to be under the user directory.

    When you then do the dataset download command, one way is to use the full LFN path to a file or directory containing files. Another way is to just provide the project name to that command and it will look for a directory with that name under `/belle/users/<user>`. With the `-u` option you can choose another user than yourself, which is the default. I set that explicitly in most commands though not really needed. Exactly how gbasf2 to figures things out internally I don't know, I just know the imperfectly documented user side of things. Here's the help string of the download command: https://pastebin.com/0P9xL0P6

2. **path regeneration**
    The path regeneration happens in the "job input sandbox". Then there's also the "job output sandbox" which contains log files from the helper functions that gbasf2 uses to execute the steering file on the grid and has a maximum size of around 10mb. But the paths are not regenerated on the grid where the datasets are stored. I still did a simple test where I added two subdirectories locally and send a job to the grid with `filename="./testdir/testsubdir/ntuple.root"` and it failed because basf2 couldn't find that directory.

3. **What I meant**
    I am not sure where exactly our misunderstanding lies. By using something like `output _path = self.get_output_file_name(basename)` the user in the examples explicitly set the output path in the `create_path()` and the path returned by `get_output_file_name()` does not depend on what is returned by `output`. Though by using `add_to_output()`, we make sure the target matches the correct output path. I had doubts about omitting the `get_output_file_name()`  part, but I thought that over and think that should be no problem.

I don't see any reason what your ideas that you mentioned under "*So the way I would handle it*" wouldn't work, the simple idea of the directories named like the root files is what I was missing. I'll think through your examples carefully on a fresh mind.

b2luigi/batch/processes/gbasf2.py
meliache: Done in 16c9d12.

b2luigi/batch/processes/gbasf2.py
meliache: In case of multiple outputs, which will each be seperately enumerated in the output dataset, I can use wildcard in the `gb2_ds_get` command (see pastebin above) to download each subset seperately, e.g. `gb2_ds_get <project>/ntupleA_*.root` and `gb2_ds_get <project>/ntupleB_*.root` into the dirs `get_output_file_name("ntuple[A|B].root")`. Or I could do my own globbing, both should be not to hard. I like your ideas.

b2luigi/batch/processes/gbasf2.py
meliache: I just recently found out that gbasf2 sometimes also automatically reschedules task. I didn't know about it until that happened to Max when testing. It seems to be documented in the Gbas2 FAQ on confluence (yeah...): Q: Are failed jobs automatically rescheduled?

> A: Jobs that failed for certain patterns are rescheduled automatically. Others are not.
> Q: How I can see if my jobs have been rescheduled?
> A: The "Job Monitor" on the DIRAC web portal does not show the counter by default, but one can choose to see it via "columns" option.

I don't know what the "certain patterns" are (I might ask today in b2questions). At least Max says that it does not reschedule all jobs that failed where it wasn't the jobs' failure, as some jobs only succeed after rescheduling manually. Gbasf2 has also a maximum number of retries/reschedules, I didn't find it anywhere documented but I grepped through the code and found a `num_reschedules=10`, so that might be it. You can see the current number of reschedules in the web gui as the last answer in the FAQ excerpt above states. I havent' found out how to find that out via the CLI.

But having such functionality in gbasf2 actually speaks against re-implementing it in the wrapper. E.g. my retry-counter doesn't count the automatic reschedules done by gbasf2. So that's a good reason to remove or at least simplify my retry code e.g. by building on the existing functionality.

I just have to think how to achieve that when the task is retried via the luigi retry policy, how to run the rescheduling command instead of `start_job`, the rest is easy.

b2luigi/batch/processes/gbasf2.py
meliache: Just had a brief look on the event handlers. Sorry for procrastinating on that, worked on some other things inbetween. So as far as I understand they operate on class level, that is that you add event handler to `Task` or subclasses of it. They are stored in the static property `_event_callbacks`. In the batch process, I have access to the task instance the event handler is operating on, but I am not sure yet how to add it only this instance (`self.task`).

I could add a generic event handler to all tasks and then put an if-condition into it which checks if there gbasf2 batch is used, e.g.

```python
@luigi.Task.event_handler(luigi.Event.SUCCESS)
def download_dataset_if_batch_is_gbasf2(task):
    if get_setting("batch", task=task) == "gbasf2":
        run_with_gbasf2(["gb2_ds_get", project_name])
```
I don't really find this solution elegant, but maybe better than side-effects in the `get_job_status`. Do you have a better idea, what do you think?

nils-braun: Ah ok. Well no, that seems not like a very elegant solution, as you have said it.

Then lets stick with your solution! We can always come back later and refine if we find a better solution.

nils-braun:
@meliache Sorry for the long pause on my side. 
I kind of lost a bit the overview: Is there anything you want my opinion on? Anything you are still working on or shall I have a last pass and then LGTM?

meliache:
It's fine, I also worked on analysis recently and let this PR stagnate a bit, just used my own branch.

> Is there anything you want my opinion on?

I wasn't sure about the retry part yet, because as I said in an [comment](https://github.com/nils-braun/b2luigi/pull/32#discussion_r416858059) I do it sub-job wise, while the built-in retry works on task-level. What would happen is that if a single sub-job in the gbasf2 job-group (project) fails, the whole task fails, luigi retries it and I could modify the `start_job()` method to reschedule failed job if it sees that the project exists already and has failed jobs. But then the retry counter would raise by one for each failed sub-job in the task. Maybe that's a good thing.

Other ideas for potential features that I had after I started this PR, but I think should be potential ideas for future PRs:

- Support basf2 variable aliases by providing an optional to add an alias dictionary. Then I could modify the steering file template that I send to the grid to apply the aliases from the dictionary on the grid. It would be cooler if I could extract the dictionary from the private member of the VariableManager that holds it, but I don't know if that's possible.

- Python modules: One potential user depends on his modules written in python, which the built-in `pickle_path` functions in basf2 don't support. I imagine it might be possible to pickle the python classes seperately via the normal `pickle.dump`, send them seperately to the grid and then re-insert them into the path, which should be possible as long as you remember the indexing. But I didn't test that idea yet and I am not sure whether that is possible.

- I realized that I also don't support non-standard conditions. Maybe there is way to enable the user to provide a function/file with additional non-path basf2 configuration functions which should be called on the grid, Or even let them define their own steering file wrapper. Or add the option to just send a steering file directly somehow, by allowing tasks which have a `steering_file` property instead of a `create_path()` method. But if I add some option like that, it should be as simple as possible so that the code stays maintainable.

nils-braun:
I am fine with keeping the "inner" job-wise retry logic. From my side the PR is fine for merging.

Your future plans look good of course, but especially the python-module-pickling is something which should be handled in basf2 and not in b2luigi.

PR #36, opened 2020-05-27
Support renamed fei file merge command for merger task

meliache:
In the PR
https://stash.desy.de/projects/B2/repos/software/pull-requests/6174/overview
the command fei_merge_files has been rename to analysis-fei-mergefiles. The old
command still works, but gives a deprecation warning. Therefore, we first try
the new command and if it doesn't exist the old command.

Alternative implementation ideas:
- Use a property function for `cmd` and put the logic which command exists into the function body.
- At the moment I only call `shutil.which` for the new command, for the old command I don't check it and just except that the subprocess call in the base class will fail later. Maybe I could check for both commands if they exist and if neither does, raise an exception. (`FileNotFoundError`?)

### DIFF
--- a/b2luigi/basf2_helper/tasks.py
+++ b/b2luigi/basf2_helper/tasks.py
@@ -1,5 +1,6 @@
 import collections
 import os
+import shutil
 
 import b2luigi
 from b2luigi.basf2_helper.targets import ROOTLocalTarget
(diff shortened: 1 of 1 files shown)


nils-braun:
Yes, I would prefer a property for `cmd`. 

I do not think you need to test both with which before. We do this in none of the other usages of cmd and in my opinion the errors by the subprocess package are understandable enough

nils-braun:
Thanks for fixing this, @meliache

PR #37, opened 2020-05-28
Update vertex fit in basf2 example to work for release 4 and new light releases

meliache:
`modularAnalysis.fitVertex` has been moved to `vertex.fitVertex` in release 4,
but that is not recommended to be used anymore. Instead, it is recommended to
use the new vertex fitter which is called `vertex.vertexTree` in release 4 and
then has been renamed to `vertex.treeFit` in the light releases following release
4 (i.e. light-2002-janus). It also performs vertex fits for all daughter decays.

The docstring of `vertex.fitVertex` says:
> Direct use of `fitVertex` is not recommended unless you know what you are doing.
> If you're unsure, you probably want to use `treeFit` or `kFit`.

So I decided to use `treeFit` with a try/except clause to enable both current versions,
depending on whether release 4 or light release of it is used. I have tested that my version of the example file no runs without failing, but didn't look at any physics plots. So I would be glad if this could be  @welschma  or @FelixMetzner could check that it makes sense physics-wise.

**Edit**: I just realized that the light releases don't include the simulation/reconstruction packages, so supporting them isn't really important as of now, but we can expect that the new function name will be used in future major releases >= 5.

### DIFF
--- a/examples/basf2/basf2_chain_example.py
+++ b/examples/basf2/basf2_chain_example.py
@@ -7,6 +7,7 @@ import basf2
 
 import modularAnalysis
 import simulation
+import vertex
 import generators
 import reconstruction
 from ROOT import Belle2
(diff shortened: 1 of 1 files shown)


nils-braun:
In my opinion the physics character of the examples is a nice to have, but it is more important that people learn how to use b2luigi.
So thanks for fixing the problem with the naming (very important), but I think we can just merge without physics plot checking

PR #38, opened 2020-05-28
Look for Y4S dec file in new location for rel 5 / master

meliache:
The old location `analysis/examples/tutorials/B2A101-Y4SEventGeneration.dec` didn't work for me when I tested the example on the current master, I checked and saw that the location was changed, so I added some code to try both the old and the new location.

Sorry for the many commits, I just wanted to test the ntuple merger task with the changes from my merge-command rename PR https://github.com/nils-braun/b2luigi/pull/36 and testing failed due to basf2 changes.

### DIFF
--- a/examples/basf2/basf2_chain_example.py
+++ b/examples/basf2/basf2_chain_example.py
@@ -7,6 +7,7 @@ import basf2
 
 import modularAnalysis
 import simulation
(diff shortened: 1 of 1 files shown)


nils-braun:
Thanks. This is in.

PR #40, opened 2020-06-22
Support aliases in gbasf2 batch

meliache:
Sam Cunliffe was so nice to tell me how to extract alias values, not just their names, from the variable manager ([b2question link](https://questions.belle2.org/question/8456/access-alias-values-in-variablemanager-from-python/). With that it was pretty easy to store them also in the pickle and reapply them on the grid. I tested it and it seems to work nicely.

I now you suggested to look at `pickable_basf2.py` to store the basf2 calls, but the variables wouldn't be pickled as far as I know except if you wrap them in `make_code_pickable`. And to be honest I couldn't figure out yet how to use the code from `pickable_basf2` for my use-case yet.

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -10,12 +10,12 @@ from collections import Counter
 from datetime import datetime, timedelta
 from functools import lru_cache
 
-from luigi.target import Target
 from b2luigi.basf2_helper.utils import get_basf2_git_hash
 from b2luigi.batch.processes import BatchProcess, JobStatus
 from b2luigi.core.settings import get_setting
 from b2luigi.core.utils import flatten_to_dict, get_log_file_dir, get_task_file_dir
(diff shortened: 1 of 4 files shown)


meliache:
The only thing that I see might be fishy with this solutions, is, that the aliases are taken from VariableManager singleton and thus are specific to the basf2 state. That is, variables added to one task will probably also be added to all tasks afterwards that are processed by the same worker, as far as I understand. But I don't see much danger with that, except when people use different alias values for the same alias names (keys) in different tasks.  I just checked that there's no bugs when people add the same alias multiple times, but then it's still only once in the variable manager, so there is no danger in that.

nils-braun:
Sorry the review took so long, looks very good to me.
I think you are very right: it should not be a problem. After all, I think (might need some checking) that basf2 tries to clean up most of its singletons after processing, including the variable manager (if not, that would be worth a PR against basf2).

Thanks, @meliache!

PR #43, opened 2020-09-29
Bugfix: Fix logic in setting gbasf2_additional_params

meliache:
The intended functionality is that the value of the setting `gbasf2_additional_params` should be added to the gbasf2
submit command string only if the setting deviates from its default value which
is `False`. I had this pattern for multiple settings and here I had introduced a
copypaste error where whether I accidentally checked for the value of the `basf2opt` setting instead of the this setting.

This can obviously lead to weird behavior, but I must admit I hadn't much used either of the settings yet. I'm surprised this went unnoticed for so long. Thanks for @welschma for finding this.


FelixMetzner (approved):
This looks sensible to me. :)

nils-braun:
Thanks you very much for the PR @meliache  and thanks to @FelixMetzner  for reviewing it

Issue #44, opened 2020-10-02
Allow dash in gbasf2 project names 


Bilokin:
Hello @meliache,

thanks a lot for the gbasf2 implementation in b2luigi, it is incredibly useful to us!
I have a small feature request: To allow "-" symbol in gbasf2 project names, that are submitted by b2luigi. 
It is an allowed symbol for BelleDIRAC and it would increase readability of project names. 

Also it would be nice to have _an option_ to not to add random string at the end of project name, but leave project name entirely up to user. The motivation is that the project names in BelleDIRAC can be only up to 32 symbols in length, and if the limit is exceeded, the project name is just truncated without any notification.

meliache:
Hi,


## 1. `-` in project names

I read the 
[gbasf2 confluence
documentation](https://confluence.desy.de/display/BI/Computing+GBasf2#ComputingGBasf2-Project)
for what is allowed in gbasf2 project names and it states:

> - The length of a project name must be less than or equal to 32 characters that consists of alphanumerics ([a-zA-Z0-9]).
>   - Allowed characters are alphanumerics ([a-zA-Z0-9]) to avoid unexpected side effect on file system or database.
>   - '.' should be avoided because it prevents registration of output to the metadata catalog, leading to job failures even when your jobs can be completed to run. See GBasf2 Troubleshooting.
>   - '-' and '_' seem to work in standard use cases, though not guaranteed
>     against every aspect.

I wanted to check the project name from within my wrapper before creating the
gbasf2 submit command, so that I can provide the user with a more helpful error
message. I wanted to only support what's "officially" allowed and that are
alphanumerics and not allow characters that might or might not work, because if
something brakes in the future, the users of the wrapper will of course ask me
for support. But if you are sure that there's no problems with that, I could
allow dashes and underscores and maybe print a warning/info message.


## 2. Option to disable the project name hash suffixes

I am aware of the 32 character limit for project names. If the user-provided
project-name-prefix is therefore longer than 22 characters (the hash is 10
characters), then b2luigi will raise an error with a message explaining the
limit. So in theory, it will either run or raise an error, nothing is truncated
without notification. So I don't really understand your motivation, maybe you
can explain it better. If you want to have the ability to have longer project
names, I can understand it, maybe I could also add a setting to shorten the
arbitrary length of 10 for the hash down to e.g. 5 or maybe even add a setting
to set it to an arbitrary length. Then, you could just set it to 0, but that
[...]

PR #45, opened 2020-10-05
Allow dashes and underscores in gbasf2 project names

meliache:
Adresses: https://github.com/nils-braun/b2luigi/issues/44

Allows `-` and `_` and gbasf2 project names, but issues a warning if any non-alphanumerics are used. These characters currently seem to work but are not officially supported by the gbasf2 devs according to this confluence quote:

> - The length of a project name must be less than or equal to 32 characters that consists of alphanumerics ([a-zA-Z0-9]).
>   - Allowed characters are alphanumerics ([a-zA-Z0-9]) to avoid unexpected side effect on file system or database.
>   - '.' should be avoided because it prevents registration of output to the metadata catalog, leading to job failures even when your jobs can be completed to run. See GBasf2 Troubleshooting.
>   - '-' and '_' seem to work in standard use cases, though not guaranteed
>     against every aspect.

In the process I changed the assert by a `ValueError` and used this PR to convert other asserts to exceptions. There, I also improved some of the value checks.

- [x] test this PR: I haven't tested the new code yet except the usual static flake8 checks of my IDE, so maybe wait with merging until I have done that. But wanted to put that here so that I can link to this PR in the issue.

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -9,6 +9,7 @@ import warnings
 from collections import Counter
 from datetime import datetime, timedelta
 from functools import lru_cache
+import re
 
 from b2luigi.basf2_helper.utils import get_basf2_git_hash
 from b2luigi.batch.processes import BatchProcess, JobStatus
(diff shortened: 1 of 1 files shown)


b2luigi/batch/processes/gbasf2.py
```diff
@@ -656,8 +664,10 @@ def get_gbasf2_project_job_status_dict(gbasf2_project_name, dirac_user=None):
     """
     if dirac_user is None:
         dirac_user = get_dirac_user()
-    assert check_project_exists(gbasf2_project_name, dirac_user), \
-        f"Project {gbasf2_project_name} doest not exist yet"
+    if not check_project_exists(gbasf2_project_name, dirac_user):
+        raise RuntimeError(
+            f"Failed to acces job status for project \"{gbasf2_project_name}\" doest not exist on the grid."
```
nils-braun, line 669: Small typo :-)

nils-braun (approved):
Thanks @meliache 
Already approving

Issue #47, opened 2020-10-09
Replace gbasf2 output string parsing with our own scripts interacting with gbasf2


meliache:
Assign this to myself, but @philiptgrace was also interested and asked about the string parsing [here](https://github.com/nils-braun/b2luigi/pull/46#issuecomment-705915640)

Interacting with gbasf2 has to be via subprocesses, there is no other way around, because `gbasf2` runs with another environment and another PYTHONPATH than b2luigi. Often, I was just lazy and called the user scripts provided by gbasf2 from the gbasf2 `BatchProcess` via subprocesses and captured and parsed their output when I needed it.

But it should be more stable to just write for each instance where I need it a short python2 script that directly imports and works with the `BelleDirac` python package which is available from the gbasf2 environment. Then I could run these scripts as subprocesses and obtain information e.g. via json or returncodes.
It is possible to just import the gbasf2 python packages and work with them.

Places where I could do that replacement of string-parsing via my own scripts
- [ ] check if dirac proxy is inititialized instead of parsing `gb2_proxy_info` [here](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L778)
- [ ] for getting the list of filenames in the output dataset instead of using `gb2_ds_list` [here](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L520)
- [ ] for checking if an output dataset even exists on the grid instead of [this](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L630)
- [ ] checking if the project already exists on the grid [here](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L680)
- [ ] for getting the dirac user [here](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L680)

PR #51, opened 2020-10-27
Use Dirac API for Dirac proxy checking and checking for gbasf2 project existance

meliache:
So far the gbasf2 batch had been parsing the output of ``gb2_job_status -p <project>`` to see if the project already has been submitted to the grid or not. But I already had a script which directly interacts with the [Dirac API](https://github.com/DIRACGrid/DIRAC/blob/integration/Interfaces/API/Dirac.py) and returns a json job status dictionary. So I now use that to also see if the project exists by return with a special exit code (currently 3) if no jobs had been found for that project name.

This is solves part of #47. 

One reason I didn't do that earlier is, because I don't fully understand the Dirac API. They seem to work mostly with return values instead of with exceptions and. And as of now I assume that if the return value of the dirac API, when querying for jobs in a project, is `status["OK"] == False`, that it means that there are no jobs for that project. But I am not sure what happens if any other exception happens on the dirac side of things, whether I will be able to distinguish between "no jobs" and a dirac error. But I just hope that dirac will actually raise an exception if something really unexpected happens.

## Description Update 2020-10-30

I forgot to mention that this PR also implements using the Dirac API for checking whether the dirac proxy is alive. This is how I started working on this branch and the project existance checking was just something I decided to add because I thought it should be simple, but when I created this PR I had first forgotten that the proxy checking is also part of this branch and somehow thought it is already merged.

## Discussion: Communication with Dirac API handler subprocesses

I use python2 scripts in subprocesses for handling the Dirac API. However, I have no well-defined way of communicating with these subprocesses yet. I wanted to go away from the  string parsing that I used with the existing gbasf2 executables. For getting complex data from subprocesses I use json. For getting boolean information I currently use exit codes. E.g. the job status getter uses both. 

However, maybe it's better to only use json for communication and not use exit codes for anything that is not an error? 

And if I keep using exit codes, I should use the allowed custom values and define them somewhere globally, e.g. in an enum. Just have to think of a way that works both in python3 and python2.


welschma (approved):
Tested the changes in this branch with gbasf2 v4r7p4, everything seems to be working again. Without the new handling of the job status, we cannot use the gbasf2 batch process withe the new gbasf2 version.

alina-ma:
Also tested it with gbasf2 v4r7p4 and it fixed the problem that the job status could not be determined, which occurred in the new gbasf2 version.

nils-braun:
So, @meliache ready to merge?

nils-braun:
Thanks to @meliache for the nice PR (as always) and thanks to @welschma and @alina-ma for checking.
@meliache I would still suggestion to make you an admin of this repo so you can take over the development (or also @welschma if he prefers). I am not sure I will be able to look into things quite often.
At least the gbasf2 part I can even check - in the core part you can if course still ask me for a review.

meliache:
Ready to merge 👍 . As always, I have ideas for things I could add (see all the issues I created for myself), but let's keep this PR as is.

I'd accept the admin request. The gbasf2 part is getting more users and sometimes it would be good to be able to react quickly to upstream changes in gbasf2 (a major release will come soon which I expect might break some workflows again). Usually, I redirect people to my fork, but it would be simpler to just have the hotfixes here (at least on a branch). And would be nice to have feature branches here instead of on my fork. I wouldn't mind having Max as a co-admin, I ask him for review/tests anyway.

Off-topic note to myself: I start to think I need to think of a testing strategy for the gbasf2 wrapper, even something rudimentary would help. So far I didn't do unit tests because it  requires a special environment with CentOS7, a gbasf2 installation and a grid certificate. But for that I could simply start creating some tests which are not part of the CI and I can run manually in the right environment whenever a new gbasf2 release comes out etc.. So far I only check whether the example task works, so I could extend on that...

meliache:
Should I maybe bump the version?

PR #54, opened 2020-12-14
Feature/remove useless super delegation

meliache:
I noticed some pylint warnings when working on features in another branch. The main thing is that the 

```python
super(<class name>, self).__init__()
```

is  not needed when no non-default arguments are passed to the constructor. The only reason why this might be useful is that new developers that aim to create their own batch-backend might start by adapting the LSF cache implementation (I did at least) and it might be good to have a reminder that they can implement a custom cache size and ttl. But even in this case, I would use the simpler python3-form super form without arguments of 

```python3
super().__init__()
```

The other change are just trailing whitespace and an explicit `check=False` in the run method, which is the default, but the zen of python (and pylint again) tell us to be specific about silencing errors. Not sure if I really agree with pylint here, but `run` is kind of new in the subprocess library and I at least always forget which is the default.

### DIFF
--- a/b2luigi/batch/processes/htcondor.py
+++ b/b2luigi/batch/processes/htcondor.py
@@ -13,9 +13,6 @@ from b2luigi.core.executable import create_executable_wrapper
 
 class HTCondorJobStatusCache(BatchJobStatusCache):
 
-    def __init__(self):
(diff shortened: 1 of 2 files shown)


nils-braun:
Looks reasonable to me! Thanks

PR #55, opened 2021-01-15
Option to only pass known command line arguments

klieret:
I want to use additional command line options in a file that uses `b2luigi.process`, but this seems to be impossible at the moment.

However, there's an easy and backwards compatible way to do this: Simply add a `ignore_additional_command_line_args` argument to the `process` function, which lets luigi simply ignore all the arguments it doesn't understand.

### DIFF
--- a/b2luigi/cli/process.py
+++ b/b2luigi/cli/process.py
@@ -89,7 +89,7 @@ def process(task_like_elements, show_output=False, dry_run=False, test=False, ba
         task_list = task_like_elements
 
     # Check the CLI arguments and run as requested
-    cli_args = get_cli_arguments()
+    cli_args = get_cli_arguments(ignore_additional_command_line_args=ignore_additional_command_line_args)
 
     if cli_args.show_output or show_output:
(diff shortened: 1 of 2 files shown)


b2luigi/cli/process.py
```diff
@@ -5,7 +5,7 @@
 __has_run_already = False
 
 
-def process(task_like_elements, show_output=False, dry_run=False, test=False, batch=False, **kwargs):
+def process(task_like_elements, show_output=False, dry_run=False, test=False, batch=False, ignore_additional_command_line_args=False, **kwargs):
```
nils-braun, line 8: I think you might also want to pass those argument to the called CLI method, or?

nils-braun (commented):
Except the one comment I have good to me. Thanks for the addition!

b2luigi/cli/process.py
klieret: Fixed, thanks! Somehow that hadn't made it into the commit.

Issue #56, opened 2021-01-27
Adapt gbasf2 batch to work gbasf2 v5 with rucio


meliache:
The output directory structure changed in release 5 of b2luigi and this needs to be incorporated in the automatic download procedure. The grid cannot be used anymore without rucio, so I don't have to think too much about backwards compatibility. I will need to do some further testing and research to see if there are other things we should change.

## Belle II confluence links about the new gbasf2

- [Gbasf2 with Rucio](https://confluence.desy.de/display/BI/GBasf2+with+Rucio) 
- [gbasf2 release v5r0p1 announcement](https://confluence.desy.de/display/BI/GBasf2+Releases#GBasf2Releases-v5r0p1)

PR #57, opened 2021-02-03
Adding sub00 to the paths

Bilokin:
Hello,

this is a PR to show my quick fix. 
It is not going to work if there are multiple `sub\d\d` directories per job, but I don't have an example of that

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -531,7 +531,7 @@ class Gbasf2Process(BatchProcess):
                 print(stdout)
                 if "No file found" in stdout:
                     raise RuntimeError(f"No output data for gbasf2 project {self.gbasf2_project_name} found.")
-                tmp_output_dir = os.path.join(tmpdir_path, self.gbasf2_project_name) + '/sub00'
+                tmp_output_dir = os.path.join(tmpdir_path, self.gbasf2_project_name, 'sub00')
                 downloaded_dataset_basenames = set(os.listdir(tmp_output_dir))
                 if output_dataset_basenames == downloaded_dataset_basenames:
                     print(f"Download of {self.gbasf2_project_name} files successful.\n"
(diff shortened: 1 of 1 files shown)


meliache:
Thanks for the PR. This should solve #56 for most cases. Did you test it and it worked?

The only remaining problem is when the job has udst and mdst-outputs, which get moved unto additional subdirectories. I did an text job which has two ntuple outputs, one mdst and one udst output and also uses the `RootOutput` module directly (with the output file `output_datastore.root`, it has the output structure:
```
/belle/user/meliache/test_new_gb2_7
    ├─ mdst
    │   ╰─ sub00
    │       ╰─ output_mdst_00000_job180712020_00.root
    ├─ sub00
    │   ├─ B_ntuple_00000_job180712020_00.root
    │   ├─ D_ntuple_00000_job180712020_00.root
    │   ╰─ output_datastore_00000_job180712020_00.root
    ╰─ udst
        ╰─ sub00
            ╰─ output_udst_00000_job180712020_00.root
```
But I think I will address that in a separate PR.

I'll quickly see if there is anything that can be improved code-wise and then I'll approve.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -531,7 +531,7 @@ def _download_dataset(self):
                 print(stdout)
                 if "No file found" in stdout:
                     raise RuntimeError(f"No output data for gbasf2 project {self.gbasf2_project_name} found.")
-                tmp_output_dir = os.path.join(tmpdir_path, self.gbasf2_project_name)
+                tmp_output_dir = os.path.join(tmpdir_path, self.gbasf2_project_name) + '/sub00'
```
meliache, line 534: Please change this to
```python
os.path.join(tmpdir_path, self.gbasf2_project_name, sub00)'
```
[path.join()](https://docs.python.org/3/library/os.path.html#os.path.join) is a special function for building path names from strings which works across operating systems and handles some special cases intelligently. In this case I think it should make no difference, but since we already have the `join` function in the code, it would be more consistent to put the `sub00` just inside that.

meliache:
I tested your updated branch on the `gbasf2_example.py` in the repo and it worked for me. Good job :+1: . I'll merge this then.

PR #63, opened 2021-02-08
Refactor away else/elif-return

meliache:
Refactored some `elif...return...`  case constucts, as after a return statement or exception there is no need else/elif. I noticed this via pylint, see [R1705](https://pheanex.github.io/pylint/#R1705). Had started refactoring this wile working on another feature branch and decided to put it into a separate PR. If you think the previous version is more readable, feel free to comment, since  according to the Zen of Python *Readability counts.* and this is the only thing that should be affected by this.


nils-braun (approved):
Yes, definitely ok with that.
In general, I prefer this style of code and I am not sure why I did not follow it in these examples. So thanks for fixing this.

If you are interested: the style with the excessive `if .. else ...` usage is called space shuttle style, and is e.g. used in k8s (which is go, so might be different anyway). But I generally do not like it that much :-)

Issue #53, opened 2020-12-14
Proposal: add Michael as the main developer on readthedocs


nils-braun:
As @meliache is now doing the main project work, he should also be added (or even replace me) on the main documentation page (https://b2luigi.readthedocs.io/en/stable/#the-team).

meliache:
Thanks for nominating me, it's fine with me if others agree.

Potential counter-arguments would be that I haven't worked much on the core of the package yet and I think you still understand the core of the code better than me. Most of my work had been on the gbasf2 wrapper that I implemented. But I guess I'll still be able to ask for your opinion on PRs even if it might take a while with you having a full-time job.

I don't think I will divert much more time to b2luigi, since I am trying to focus my efforts on my PhD topic for my last year. But I guess this is not what you had been asking for. At least, I have more contact to the user-base than you have.

Cheers
Michael

PR #65, opened 2021-02-08
Update team list with Michael as main developer

meliache:
The main change introduced by this commit is that following the proposal in #53 I added myself as the main developer, but added for @nils-braun the special position of _Original author_.

I checked the contributors from the git logs and added @klieret , @bilokin and @philiptgrace, who all helped out, but were not in the list yet. Further I added the links to the github user pages for all contributors for whom I could find them, i.e. in addition to the ones I already mentioned also @eckerpatrick, @welschma, @FelixMetzner.

I @-mention you people so that you can object in the comments if you have any privacy concerns or similar about your real name and github profile being linked here. 

The last change introduced in this commit is that I slightly improved the formatting by using an RST definition list, so now it looks like this:
grafik

If wanted, I could also put in some work to add the github user thumbnails next to the author names, as e.g. @klieret  had done [here](https://github.com/hsf-training/hsf-training-ssh-webpage/blob/gh-pages/README.md#authors).

But I shouldn't distract too much from the main change, me as the main developer, so comment e.g. if you want the position to be called differently. After all, 80% of the commits are still from Nils.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -126,12 +126,12 @@ Original author:
 
 Features, fixing, help and testing:
 
-* Felix Metzner
-* Patrick Ecker
+* Felix Metzner (`FelixMetzner`_)
+* Patrick Ecker (`eckerpatrick`_)
 * Jochen Gemmler
-* Maximilian Welsch
-* Kilian Lieret
-* Sviatoslav Bilokin
+* Maximilian Welsch (`welschma`_)
+* Kilian Lieret (`klieret`_)
+* Sviatoslav Bilokin (`bilokin`_)
 * Phil Grace
 
 Stolen ideas:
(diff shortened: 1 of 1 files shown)


meliache:
Not sure if I should replace @nils-braun  with myself also in the [pyproject.toml](https://github.com/nils-braun/b2luigi/blob/main/pyproject.toml#L7) (where Nils still has an kit.edu E-Mail) and the [docs/conf.py](https://github.com/nils-braun/b2luigi/blob/main/docs/conf.py#L24).

nils-braun (approved):
Thanks @meliache - I am fine with the changes. 
Thanks again for all the hard work you put into the project, you definitely deserved being the main author here!

Concerning the other config files - you can proceed as you wish. I have no problem being mentioned there (as, after all, I have never received any request other than on github), but if you think that some issues might not come back to you, feel free to swap in your name.

(and yes, my mail address is of course nonsense now...)

Issue #69, opened 2021-02-11
Continue running other tasks when one gbasf2 download failed


meliache:
The `Gbasf2Process` marks a luigi task that is associated with a gbasf2 project as `successful` when all the jobs in that project are `DONE`. However, what then remains is downloading the outputs of those jobs and there it often happens that `gb2_ds_get` doesn't manage to download all files. Currently, `luigi` then validates the download, sees that the downloaded directory is missing files and then instead of moving the downloaded dataset from the partial download directory to the final output directy, it raises a `RuntimeError` that the download wasn't successful.

But that exception stops the whole luigi process and the processing of all tasks. It would be more conventient to let the processing of all other tasks/projects continue in such a case.

That could be solve by just printing out a warning and doing an early return in the download function so that the partial download is never moved to the output directory.

However, I must check that the process isn't at any point marked as successful just because the `get_job_status()` method returns `successful`. However, it is currently coupled to the status on the grid and not to the download status, as I use it to decide when to do the download. Here I'll have to think how to solve that cleverly.

PR #72, opened 2021-02-16
Don't raise exception when download failed

meliache:
Instead don't produce the outputs. WIP goal is to also mark the job as failed during runtime. Addresses #69

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -286,8 +286,15 @@ class Gbasf2Process(BatchProcess):
         if n_done == n_jobs:
             # download dataset only the first time that we return JobStatus.successful
             if not self._project_had_been_successful:
-                self._on_first_success_action()
-                self._project_had_been_successful = True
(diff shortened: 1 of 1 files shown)


meliache:
Okay the solution was easier than I though, since I initiate the download in `get_job_status` anyway, I just catch the error and return `JobStatus.aborted`.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -296,8 +286,15 @@ def get_job_status(self):
         if n_done == n_jobs:
             # download dataset only the first time that we return JobStatus.successful
             if not self._project_had_been_successful:
-                self._on_first_success_action()
-                self._project_had_been_successful = True
+                try:
+                    self._on_first_success_action()
+                    self._project_had_been_successful = True
+                # RuntimeError might occur when download out output dataset was not complete. This is
```
meliache, line 292:
```suggestion
                # RuntimeError might occur when download of output dataset was not complete. This is
```

PR #75, opened 2021-02-25
Add a setting to provide a text file with grid LFNs

meliache:
Such text files are now used by many analysts, as you get them from the grid dataset searcher and they allow you to run over multiple datasets in a single project.


ArturAkh:
What is the estimate, when this will be included into the main branch?

Is there any reason to wait?

P.S.: This would make #90 obsolete.

meliache:
Okay I know added some unittests for the function `Gbasf2Process._build_gbasf2_submit_command()`  where I change the task parameters and check if the resulting gbasf2 project submit command looks how we expect. So I'm quite certain now I avoided any stupid logic mistakes. 

However, I haven't tested this in practice yet. However, the alternative PR #90 by  results in the same command, so if @ArturAkh tells me that he tested his version and it worked, I think we can merge this without worrying.

ArturAkh:
@meliache fine with me!, I'll anyhow will test it on my usecase in

https://github.com/ArturAkh/b2luigi/tree/fei_with_gbasf2

as soon the feature is merged and will tell you, in case I find something not working

ArturAkh:
What is your time estimate to merge that change? I'd like to have it rather sooner than later :) I have some other changes in the pipeline to put on master...

PR #76, opened 2021-03-11
Adding an option to set the lsf job_name parameter

MarcelHoh:
I find it can be useful to get a quick overview of the running jobs through their job_name parameters on kekcc. With b2luigi submitted jobs this is set to "{FULL_PATH}/executable_wrapper.sh" which is not easy to parse at a glance. 

This PR adds the option to set a 'job_name' attribute for your tasks which is then submitted as the lsf job name (-J task.job_name).

### DIFF
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -42,6 +42,14 @@ class LSFProcess(BatchProcess):
 
       The default is the short queue "s".
 
+    * the LSF job name can be contralled via the ``job_name`` parameter, e.g.
+      .. code-block:: python
+
+        class MyLongTask(b2luigi.Task):
+            job_name = "my_fun_job"
+
+      The default "{FULL_PATH}/executable_wrapper.sh".
+
     * By default, the environment variables from the scheduler are copied to
       the workers.
       This also applies we start in the same working directory and can reuse
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -78,6 +78,11 @@ class LSFProcess(BatchProcess):
         except AttributeError:
             pass
 
+        try:
+            command += ["-J", self.task.job_name]
+        except AttributeError:
+            pass
+
         log_file_dir = get_log_file_dir(self.task)
         os.makedirs(log_file_dir, exist_ok=True)
 
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -31,12 +31,12 @@ class LSFProcess(BatchProcess):
     Reference implementation of the batch process for a LSF batch system.
 
     Additional to the basic batch setup (see :ref:`batch-label`), there are
-    LSF-specific settings. These are:
+    LSF-specific :meth:`settings <b2luigi.set_setting>`. These are:
 
     * the LSF queue: ``queue``.
     * the LSF job name: ``job_name``.
 
-    Both can be controlled via the standard settings interface, e.g.
+    For example:
 
       .. code-block:: python
 
@@ -44,8 +44,11 @@ class LSFProcess(BatchProcess):
             queue = "l"
             job_name = "my_long_task"
 
-      The default queue is the short queue "s". If no job_name is set the task
-      will appear as "{result_dir}/parameter1=value/.../executable_wrapper.sh"
+      The default queue is the short queue ``"s"``. If no ``job_name`` is set the task
+      will appear as ::
+
+        <result_dir>/parameter1=value/.../executable_wrapper.sh"
+
       when running ``bjobs``.
 
     * By default, the environment variables from the scheduler are copied to
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -79,8 +79,9 @@ class LSFProcess(BatchProcess):
     def start_job(self):
         command = ["bsub", "-env all"]
 
-        queue = get_setting("queue", task=self.task, default='s')
-        command += ["-q", queue]
+        queue = get_setting("queue", task=self.task, default=None)
+        if queue is not None:
+            command += ["-q", queue]
 
         job_name = get_setting("job_name", task=self.task, default=None)
         if job_name is not None:
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -30,25 +30,23 @@ class LSFProcess(BatchProcess):
     """
     Reference implementation of the batch process for a LSF batch system.
 
-    Additional to the basic batch setup (see :ref:`batch-label`), additional
-    LSF-specific things are:
+    Additional to the basic batch setup (see :ref:`batch-label`), there are
+    LSF-specific settings. These are:
 
-    * the LSF queue can be controlled via the ``queue`` parameter, e.g.
+    * the LSF queue: ``queue``.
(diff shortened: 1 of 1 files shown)


meliache:
Hi, thanks for the PR, I'm always happy if I see that people contribute. I'll give this a proper reviewer once I find some free time. I basically inherited this project from Nils, but haven't used the LSF batch much and actually never used it directly via the CLI.  First of all thanks for including documentation right away.

You expect the job name to be provided as a parameter of the task class/object. I'm thinking (but yet undecided) whether it wouldn't be better to use the b2luigi settings interface that is used for all other settings, i.e. something similar to

``` python
job_name = b2luigi.get_setting("job_name", task=self.task,  default=None)
if job_name is not None:
    command += ...
 ```
 
This would still allow setting the job name via a task parameter, but would also allow using `b2luigi.set_setting()` and providing the setting via a `settings.json` file.
 
My question: Does the LSF job name have to be unique per task? E.g. is it possible to submit multiple tasks with the same job name? If that is not possible, then providing a global job name via `set_setting()` or `settings.json` might be unwanted and there is a valid reason to only allow setting it inside tasks. However, if it needs to be unique, then the code in your documentation example is not good, since it would use the same job name for all instances (with different luigi parameters) of a task class. So I'm not sure which is the case.

**Comment by MarcelHoh** [comment]
Hi, thanks for the quick response! For lsf the job_name does not need to be unique. There is a separate unique id assigned when the job is submitted.

I have no strong opinion as to how to best implement the parameter, for now I just followed the approach taken by the ```queue``` parameter. Currently if I want to adjust the job_name for a particular instance of the class I overwrite the attribute in the ```__init__``` function but I think this should still be possible if it is switched to using the settings interface.

**Comment by meliache** [comment]
>  I just followed the approach taken by the queue parameter
Ah I just saw that, I now checked out your branch on my notebook to take a closer look. Then its consistent and I'm fine with it, though I would be interested in hearing from @nils-braun if there's a specific reason the `queue` is not implemented as a setting.

File: b2luigi/batch/processes/lsf.py [hunk 1]
```diff
@@ -42,6 +42,14 @@ class MyLongTask(b2luigi.Task):
 
       The default is the short queue "s".
 
+    * the LSF job name can be contralled via the ``job_name`` parameter, e.g.
```
meliache, line 45: You need an empty newline before `.. code-block`, otherwise the formatting is broken and when I build the documentation it looks like this:
image

```suggestion
    * the LSF job name can be contralled via the ``job_name`` parameter, e.g.
    
```

You can check by yourself by building it by running `make html` in the `docs` directory and then opening the file `_build/index.html` in your web-browser.

- [ ] fix documentation formatting

nils-braun:
No reason, other than I probably introduced the queue parameter before I started with the settings (but maybe I am wrong here) :-)

Would be nice to also move the queue parameter - I would vote to have both be accessible via settings.

b2luigi/batch/processes/lsf.py
```diff
@@ -73,10 +79,12 @@ def get_job_status(self):
     def start_job(self):
         command = ["bsub", "-env all"]
 
-        try:
-            command += ["-q", self.task.queue]
-        except AttributeError:
-            pass
+        queue = get_setting("queue", task=self.task, default='s')
```
nils-braun, line 82: Was ist intended to have a default here? While "s" might be the best default setting for the KEK batch system, that is not something "universal" (and e.g. people running on other batch farms will need to change that). I would love to make this part as general as possible.

MarcelHoh: good point! I will fix that. I think for lsf it should still default to 's' anyway. I'm not familiar with the other batch systems, do they have a 'job_name' parameter or something analogous?

meliache: Concerning your question, I just searched the documentation for `condor_submit` and it has a 
`-batch-name` setting which seems to serve an identical purpose to the lsf `job_name`. This could be handled by the same setting.

Gbasf2 projects have a project name, which for gbasf2 is required, since for gbas2 you usually submit a steering file and an input dataset and gbasf2 handles the creation of individual jobs for that dataset, grouped under that project name. The b2luigi gbasf2 wrapper that I wrote is therefore different from the htcondor or lsf BatchProcess'es, as it treats the whole project (group of jobs) as an individual job. In that case I had to make the project name a required setting and also make sure that it is unique, so I ask the users to provide a `gbasf2_project_name_prefix` setting and automatically add a unique hash as a suffix to the project name.

b2luigi/batch/processes/lsf.py
```diff
@@ -30,17 +30,23 @@ class LSFProcess(BatchProcess):
...
       .. code-block:: python
 
         class MyLongTask(b2luigi.Task):
             queue = "l"
+            job_name = "my_long_task"
 
-      The default is the short queue "s".
+      The default queue is the short queue "s". If no job_name is set the task
+      will appear as "{result_dir}/parameter1=value/.../executable_wrapper.sh"
+      when running ``bjobs``.
```
meliache, line 49: I would put the default parameters and variable names in literal blocks. Also in documentation I usually know the notation to use angle brackets for strings to be substituted instead of curly brackets, though the latter are used to python f-strings and templating systems. I'm not sure but I think the angle-brackets notation originates from the [Backus-Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form). My suggestion:


```suggestion
      The default queue is the short queue ``"s"``. If no ``job_name`` is set the task
      will appear as ::

        <result_dir>/parameter1=value/.../executable_wrapper.sh"

      when running ``bjobs``.
```

b2luigi/batch/processes/lsf.py
```diff
@@ -30,17 +30,23 @@ class LSFProcess(BatchProcess):
...
-    Additional to the basic batch setup (see :ref:`batch-label`), additional
-    LSF-specific things are:
+    Additional to the basic batch setup (see :ref:`batch-label`), there are
+    LSF-specific settings. These are:
 
-    * the LSF queue can be controlled via the ``queue`` parameter, e.g.
+    * the LSF queue: ``queue``.
+    * the LSF job name: ``job_name``.
+
+    Both can be controlled via the standard settings interface, e.g.
```
meliache, line 39: I think it's not necessary to repeat that these can be controlled via settings, because in line 34 you already mention that these are settings and in the b2luigi documentation it should be clear that b2luigi settings are meant. So I would remove this sentence or replace it by a simple _For example:_
```suggestion
    For example:
```

b2luigi/batch/processes/lsf.py
```diff
@@ -30,17 +30,23 @@ class LSFProcess(BatchProcess):
     """
     Reference implementation of the batch process for a LSF batch system.
 
-    Additional to the basic batch setup (see :ref:`batch-label`), additional
-    LSF-specific things are:
+    Additional to the basic batch setup (see :ref:`batch-label`), there are
+    LSF-specific settings. These are:
```
meliache, line 34: Here we can make "settings" a link to the settings documentation and then we don't need to specify below which settings interface we mean.

```suggestion
    LSF-specific :meth:`settings <b2luigi.set_setting>`. These are:
```

meliache:
Fixed the docstring formatting and created #79 for using the `job_name` setting for `htcondor` too, now I think this is good to merge :)

MarcelHoh:
sorry I was a little busy over the last few days. Thank you very much for fixing it up and merging it!

Issue #35, opened 2020-05-27
Maybe support non-path basf2 configuration in gbasf2/grid batch


meliache:
I just sent the basf2 path to the grid, but sometimes the user might want to call some basf2 configuration functions which change the internal state of basf2, but are not stored in the path, and they might not be available via CLI arguments. Examples include

- basf2 aliases (stored in the variable manager)
- testing payloads
- conditions

In principle I could let the user add all these things via additional settings. E.g. let the user provide a dictionary with all the aliases and a list with the testing payloads, then in the steering file wrapper I can add the aliases and the testing payloads. But it's hard to anticipate everything the user might want and it might bloat my code.

It would be nice if the user could provide a function or a file in which all basf2 configuration functions, that the user needs, are called. I'm not sure however how to call that from the steering file wrapper. I could send an additional basf2 configuration file to the grid, but not sure how I could import that configuration from the steering file wrapper, without forcing the user to supply a specific function name.

Another possiblity would be to have the user supply his own steering file template. Code-wise this seems the simplest solution, so I already did a test-implementation on [a branch](https://github.com/meliache/b2luigi/tree/feature/support-custom-gbasf2-steering-file-wrapper-template), but it is dangerous and would require careful documentation, here is my first attempt
![Bildschirmfoto vom 2020-05-27 14-36-51](https://user-images.githubusercontent.com/5121824/83019712-b2abe580-a027-11ea-804b-04030d623d87.png)
But I am not sure I like this solution, it's not something normal users should have to use for some simple configuration.

I would be happy for some suggestions if you have any ideas how to support basf2 configuration without adding much complexity and in a simple way for the user. I would be happy to do implement that myself.

nils-braun:
I will look into this soon, but are you aware of the `pickable_basf2` module that tkeck wrote (framework/scripts). I do not know if this is still working or in use, but I guess if something is already supported by basf2, we should use that.

meliache:
Niceeeee, looks very good. Wasn't aware of when writing this, might have heard. Will definetly test that. Might have found that earlier when reading more carefully the code of `pickle_path.py`, because the function `make_code_pickable()` mentions that.

I didn't create this issue for necessarily for you to invest time looking into that, just wanted to have that as a TODO basically assigned to myself, but where I would be happy for outward input. This was one such example. But it's not urgent.

meliache:
[...]

PR #77, opened 2021-03-17
Include global tags in pickled path object

philiptgrace:
Partially addresses #35 by saving the global tags list to the pickle file.

### DIFF
--- a/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
+++ b/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
@@ -1,5 +1,6 @@
 import pickle
 
+from basf2 import conditions as b2conditions
 from basf2.pickle_path import serialize_path
 from variables import variables as vm
 
@@ -28,4 +29,5 @@ def write_path_and_aliases_to_file(basf2_path, file_path):
     with open(file_path, 'bw') as pickle_file:
         serialized = serialize_path(basf2_path)
         serialized["aliases"] = get_alias_dict_from_variable_manager()
+        serialized["globaltags"] = b2conditions.globaltags
         pickle.dump(serialized, pickle_file)
--- a/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
+++ b/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
@@ -16,7 +16,7 @@ def get_alias_dict_from_variable_manager():
 
 def write_path_and_state_to_file(basf2_path, file_path):
     """
-    Serialize basf2 path and variables from variable manage to file.
+    Serialize basf2 path and variables from variable manager to file.
 
     Variant of ``basf2.pickle_path.write_path_to_file``, only with additional
     serialization of the basf2 variable aliases and global tags.
--- a/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
+++ b/b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
@@ -14,14 +14,16 @@ def get_alias_dict_from_variable_manager():
     return alias_dictionary
 
 
-def write_path_and_aliases_to_file(basf2_path, file_path):
+def write_path_and_state_to_file(basf2_path, file_path):
     """
-    Serialize basf2 path and variables from variable manage to file
+    Serialize basf2 path and variables from variable manage to file.
 
     Variant of ``basf2.pickle_path.write_path_to_file``, only with additional
-    serialization of the basf2 variable aliases.  The aliases are extracted from
-    the current state of the variable manager singleton and thus have to be
-    added in the python/basf2 process before calling this function.
+    serialization of the basf2 variable aliases and global tags.
+
+    The aliases are extracted from the current state of the variable manager singleton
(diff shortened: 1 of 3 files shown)


philiptgrace:
Tested this works as expected using the following two scripts.

```python
## 01-write-pickle.py

import basf2 as b2
import modularAnalysis as ma

from b2luigi.batch.processes.gbasf2_utils.pickle_utils import (
    write_path_and_aliases_to_file,
)

files = ["/group/belle2/dataprod/MC/SkimTraining/proc11_exp10.mdst.root"]
pickle_file_path = "load_gammas.pickle"

path = b2.Path()
ma.inputMdstList("default", files, path=path)

ma.fillParticleList("gamma:NoSelections", "", True, path)
ma.variablesToNtuple(
    "gamma:NoSelections", ["p", "E"], filename="load_gammas.ntuple.root", path=path
)
write_path_and_aliases_to_file(path, pickle_file_path)

print(f"Global tags: {b2.conditions.globaltags}")
```

```python
## 02-read-pickle.py

import pickle

import basf2 as b2
from basf2 import pickle_path as b2pp


def get_global_tags_from_file(file_path):
    """
    Extract list of global tags from pickle file, and update the list of global tags.
    """
    with open(file_path, "br") as pickle_file:
        serialized = pickle.load(pickle_file)
    try:
        b2.conditions.globaltags += serialized["globaltags"]
    except KeyError:
        pass


pickle_file_path = "load_gammas.pickle"

get_global_tags_from_file(pickle_file_path)
print(f"Global tags: {b2.conditions.globaltags}")
path = b2pp.get_path_from_file(pickle_file_path)

b2.print_path(path)
b2.process(path, max_event=100)
```

b2luigi/batch/processes/gbasf2_utils/pickle_utils.py
```diff
@@ -28,4 +29,5 @@ def write_path_and_aliases_to_file(basf2_path, file_path):
     with open(file_path, 'bw') as pickle_file:
         serialized = serialize_path(basf2_path)
         serialized["aliases"] = get_alias_dict_from_variable_manager()
+        serialized["globaltags"] = b2conditions.globaltags
```
meliache, line 32: The function name `write_path_and_aliases_to_file` and its description don't reflect that globaltags are also saved. To the user of the function it should be obvious what's happening. So I would either:
- add the globaltags outside this function, possibly via a separate function
- What I prefer would be to change the function name to something more general like `write_path_and_state_to_file()` and document in the docstring that aliases and basf2 paths are also stored. Here we don't have to worry about breaking backwards compatibility, since this we are still in the `0.` release numbers. 

One could alternatively add also function parameters (`save_aliases=True, save_global_tag=True`) to steer which parts of the global states are stored, but I don't think we really need this

- [ ] reflect global tag addition in function name and docs

meliache:
Looks good to me :+1: 

The only issue I found you can find in the comment above. Not sure if we should document this elsewhere, I don't explicitly say anywhere that globaltags are not supported, it is only implicit in the [caveats](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/gbasf2.py#L73) section of the documentation where I write
> - It can be used **only for pickable basf2 paths**, as it stores
          the path created by ``create_path`` in a python pickle file and runs that on the grid.

philiptgrace:
This PR only implements one part of the `ConditionsConfiguration` object. If the `BASF2StateRecorder` solution in #35 doesn't work out, then we could capture more of the state by making the entire `ConditionsConfiguration` object pickle-able. From some quick testing, it seems we'd need to make a PR to basf2 to make this this work.

meliache:
@philiptgrace Thanks for the quick fix of the naming/docs, if the checks from our new CI succeed I'll merge this.
 
> If the `BASF2StateRecorder` solution in #35 doesn't work out, then we could capture more of the state by making the entire `ConditionsConfiguration` object pickle-able. From some quick testing, it seems we'd need to make a PR to basf2 to make this this work.

Sounds like a good idea but something for a separate PR. I remember that didn't get the solution with the `BASF2StateRecorder` from `pickable_basf2` to work for our use-case, but it might be just me :man_shrugging:. I gave up on that but feel free to try.

If you figure out how to pickle the entire `ConditionsConfiguration` I'd be happy. We can always try to add more and more code to pickle different parts of the basf2 state, but if possible I'd try to keep things simple ([KISS](https://en.wikipedia.org/wiki/KISS_principle)).

I'm still not sure whether it wouldn't be a simpler solution when we would allow the user to provide an additional python file that will then also be send to the grid via `gbasf2 -f` and that can be imported from the steering file wrapper (the jinja template). So the user could have a file `grid_basf2_setup.py` with content like:

``` python
import basf2
basf2.conditions = ...
...
```

And then from the steering file wrapper we could run
``` python
try:
    import grid_basf2_setup
except ModuleNotFoundError:
    pass
```

Not sure if this would work and even if so, it might  lead to various kinds of problems that I can't predict now, so just pickling parts of the basf2 state seperately seemed easier. But if we ever really need something and this can't be just added in any other way, this or something similar might be an approach.

Issue #74, opened 2021-02-18
Automate deployment


meliache:
Automate the bumpversion, push and publishing to PyPi, which are currently separate steps as described in the [development documentation](https://b2luigi.readthedocs.io/en/stable/advanced/development.html#local-development). It previously contained the comment by @nils-braun

> At a later stage, I will try to automate this.

But as far as I know this had not been implemented it. I removed that comment in the development PR #73  and decided to turn it into this issue here for more visibility and for keeping the docs up-to-date.

daritter:
Speaking of this: Can I have a new release on pypi? The last version on pypi depends on luigi<2.7.9 but 2.7.8 is not compatible with a recent jupyter due to conflicting tornado requirements.

The git repository has updated requirements but there's no release for this.

meliache:
> Speaking of this: Can I have a new release on pypi?

@daritter Thanks for the reminder. The tornado issue is why we updated the required luigi version in #52. I'll try to do a minor version bump this week (might wait for PRs #76 and #77 which are close to being merged) and then ask @nils-braun to publish this to PyPi since I don't have admin rights there. 

As a new github maintainer I wanted to update the version since a while but kept postponing that b/c there were always new PR's and issues, especially with the new gbasf2 version which I tried to keep up with. But I think I should just do the version bump at some arbitrary point now, we don't have as strict a release schedule as you with basf2 and I can just do a new patch release any short time later.

Then, I guess you might be interested into updating the version in the basf2 externals. I'd say that the local/LSF/htcondor batches are pretty stable and should be updated (not sure if the htcondor one is even in the externals release), for any user of the gbasf2 batch I'd recommend to install the latest release from github to keep up with braking changes in gbasf2.

daritter:
b2luigi is AFAIK not used in production except for the systematics framework and they depend on gbasf2.

So we can remove b2luigi from the externals hand have users install it manually. But if there's a version that works with the systematic framework which can be included in the externals that would of course be easier.

nils-braun:
@meliache I can make you an author of the pipy repo (what is your pipy username?). But we might also want to fix this issue right away. I will see if I can find some time today evening.

meliache:
@nils My PyPi username `meliache`, same as on github, haven't much used that though.

[...]

PR #78, opened 2021-03-17
Added github workflows and removed travis

nils-braun:
Solves #74 (except the bumpversion, which can not be solved as we are not using git flow).

So far, whenever there is a new tag, there is in principle also a pipy publish via Travis. However, Travis is a bit flaky sometimes and GitHub actions offer more possibilities anyhow, so let's move to GitHub actions :-)

This PR removes the Travis integration and adds two GitHub actions: one for publishing (whenever a new GitHub release is created, which typically should be done on git tags) and one for the usual tests of PRs. I also moved to codecov in this PR (just because it is nicer). 

This can be expanded to also include things like stylechecks (pycodestyle, autopep8, black), isort, flake8 etc - but so far letÄs start with the basics.


nils-braun:
@meliache Feel free to either merge it if you like to comment on it. That is of course only my own proposal

.github/workflows/test.yml
```diff
@@ -0,0 +1,56 @@
...
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
meliache, line 19: Just fyi, as you can't know, but the current basf2 externals still use python 3.6.10, so the code should work with that, but I don't think it's an issue, I just [checked](https://docs.python.org/3.7/whatsnew/3.7.html)  there is no new syntax in 3.7 (walrus only appears in 3.8)

meliache:
Thanks, looks good to me :slightly_smiling_face:. And if you agree with my one comment above that testing with 3.7 is probably no problem, then I'll happily merge.

.github/workflows/test.yml
nils-braun: We can also check with both 3.6 and 3.7.
Let's do this in another PR (at a later point in time)

PR #79, opened 2021-03-18
Make `job_name` setting set the `-batch_name` for htcondor

meliache:
In #76 the `job_name` setting was introduced to optionally set the `job_name` parameter for LSF jobs, to give groups of jobs a meaningful name for easier checking the status of specific jobs via the CLI. HTCondor has an equivalent `JobBatchName` setting which can already in previous versions be set via the `htcondor_settings` dictionary, but for consistency, this PR also makes htcondor observe the existing `job_name` setting.

## TODO's:
- [x] I just added the code, which was straightforward, but still need to run tests to see if it actually works.
- [x] Once I test the code, I also want to check if jobs get an default value for the job batch name automatically. If not, I could set this by default to the task class name, task id or the path to the executable wrapper, which seems to be the default for LSF jobs.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.1
+current_version = 0.6.0
 commit = True
 tag = True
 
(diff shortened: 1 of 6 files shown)


meliache:
Because `codecov` complained that the diff is less than 50% covered by tests, I added unittests to `_create_htcondor_submit_file`, so this PR grew much larger than expected, but at least I learned how to test `BatchProcess` class methods without having to instantiate the whole `BatchProcess` class, it would be easy in the future to test helper functions of other batch classes.

PR #82, opened 2021-03-22
Install flit dependency in deploy workflow

meliache:
Deployment action for release v0.6.0 [failed](https://github.com/nils-braun/b2luigi/runs/2165467099?check_suite_focus=true) because of this.


nils-braun:
Thanks @meliache

PR #83, opened 2021-03-24
import get_setting for lsf tasks

MarcelHoh:
Hi, when we swtiched to get_setting for the ``queue`` and ``job_name`` parameter we did not add an import to use get_setting.

There is an extra problem in that for the ``queue`` and ``job_name`` parameter the default value is set to ``None``. However get_setting internally takes ``None`` to mean the setting is not defined and throws a ValueError. Should this be removed to allow for the default value to be ``None``?

### DIFF
--- a/b2luigi/batch/processes/lsf.py
+++ b/b2luigi/batch/processes/lsf.py
@@ -7,7 +7,7 @@ from b2luigi.batch.processes import BatchProcess, JobStatus
 from b2luigi.batch.cache import BatchJobStatusCache
 from b2luigi.core.utils import get_log_file_dir
 from b2luigi.core.executable import create_executable_wrapper
-
+from b2luigi.core.settings import get_setting
 
 class LSFJobStatusCache(BatchJobStatusCache):
 


MarcelHoh:
Woops just saw the get_settings import has already been fixed.

meliache:
@MarcelHoh Still thanks for reporting. Last week I was a bit impatient with PR's because I wanted to merge most open PR's before publishing a new [release](https://github.com/nils-braun/b2luigi/releases) to make the latest version available on PyPi, i.e. so that you can just do `pip3 install --upgrade b2luigi` instead of having to install the development version, e.g. via `pip3 install --upgrade "git+https://github.com/nils-braun/b2luigi"`. 

I had somehow assumed you had already tested the new settings code in your PR, but really, I should have tested it myself. I saw the error when implementing the `job_name` setting for `htcondor`, where the same error appeared. Actually, I noticed the missing import right after opening the file in emacs, because my syntax checker underlined that line in red (but any IDE would have done so). This just shows that we should add static syntax checking (e.g. `flake8`) to our github actions.

Anyway, I fixed it in #81 if I had the time I would have made you reviewer, but somehow I really wanted to get done with the new release on the weekend. Btw, thanks to you I just noticed that I have mistakenly not included that PR in v0.6.0, so I just published a new patch release v0.6.1

Tipp: You can use the `watch` button on github to get email notifications for new releases or PR's and Issues. You can customize via the dropdown menu what events you'll get notifications for.

PR #86, opened 2021-03-24
Feature/add pre commit linters

### LINKED ISSUE #85: Add static linter to workflows
At least for error-detection.

In #78 @nils-braun added github workflows and mentioned

> This can be expanded to also include things like stylechecks (pycodestyle, autopep8, black), isort, flake8 etc - but so far let's start with the basics.

I think what can't hurt in any way is to at least add static syntax checking for errors with `pyflakes` (the error-part of `flake8`) or the error-detection of `pylint`. I'm not against going the extra step to do style checks, but that is a bit more work since it involves discussing which style we actually want (e.g. line length), defining it in a config file and also mentioning it in the documentation, so imo this could be split into a separate PR. 

I have little experience with github workflows, but personally I have used [pre-commit](https://pre-commit.com) to do pre-commit checks locally. There's also a [pre-commit action](https://github.com/marketplace/actions/pre-commit) in the github marketplace and that would be my first idea how to implement style checks, since I find that extendible and straightforward and would feel confident of adding that myself. I just saw that there are also existing actions for e.g. flake8 that annotate the code, but I don't think we really need that. Or we could write our own actions. Don't know what's the most common approach.

meliache:
Add pre-commit linters and flake8-checks for PEP8.

Didn't to much configuration for  flake8 except except a very loose max line length of 132, which basf2 uses. Personally I usually try to stay somewhere below 120, so feel free to suggest something smaller. Also we could add custom ignores for certain errors or classes of errors but usually, e.g. pycodestyle warnings (`W`), but I think the defaults should be fine for a start and we can add exceptions later if we need them.

I just included the pre-commit checks in the test-action for the moment, but I could also create a custom github action with a custom yaml-file for that, but I fear that would result in some duplicated yaml-code, not sure what's best there. It fits very loosely the definition of a test. The pre-commit configuration is adapted from [here](https://github.com/pre-commit/action#using-this-action). I'm not sure if it requires pre-commit to be installed as a depencency, I haven't tested the action yet.

Also I updated the developer documentation on how to installl pre-commit and while I was at it I also changed the publishing part to mention that it's now done automatically via github actions.

Fixes #85


meliache:
Seems the pre-commit action already runs for this PR and it was checking all files and not only the changed ones, so I now commited all pre-commit suggestions. Most are trailing-whitespace, end-of-file newlines, etc. So if viewing the diff, I recommend disabling whitespace diffs, which can be done in github in the diff settings (gear icon) or in the CLI with

``` python
git diff --ignore-all-space --ignore-blank-lines main...feature/add-pre-commit-linters 
```

nils-braun (approved):
Looks good to me!

PR #88, opened 2021-03-30
Fixed incorrect call to warnings.warn

philiptgrace:
Calling with `RuntimeError` raises a `TypeError`


meliache:
Thanks, this was a bad error :see_no_evil:

PR #90, opened 2021-04-14
b2luigi/batch/processes/gbasf2.py: providing inputs as .txt file

ArturAkh:
Allow the possibility to provide input datasets to `gbasf` listed in a `.txt` file (assuming this extension for simplicity)

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -410,7 +410,11 @@ class Gbasf2Process(BatchProcess):
 
         gbasf2_input_dataset = get_setting("gbasf2_input_dataset", default=False, task=self.task)
         if gbasf2_input_dataset is not False:
-            gbasf2_command_str += f" -i {gbasf2_input_dataset} "
+            if ".txt" not in gbasf2_input_dataset:
+                gbasf2_command_str += f" -i {gbasf2_input_dataset} "
+            # Assuming, that instead of a local or LFN path, a .txt file with these paths is provided
+            else:
+                gbasf2_command_str += f" --input_dslist {gbasf2_input_dataset} "
 
         gbasf2_n_repition_jobs = get_setting("gbasf2_n_repition_job", default=False, task=self.task)
         if gbasf2_n_repition_jobs is not False:


meliache:
This PR duplicates PR #75, so we have to decide which we want. The reason why I didn't merge #75 is simply because I didn't find any time to test it yet. My design choice was using a second setting and basically do an xor check that one of those two settings must be set, but not both. If I didn't do a stupid mistake in that it should work, but I didn't want to merge it without testing.

I assume you have tested this PR, right?

You're idea of reusing the existing setting is also nice. I have to think which is clearer to the user.

Please see issue #75 for the feature request and some existing discussion (by me).

Anyway, if I use your PR, please add some documentation for the behaviour of the setting.

Concerning the code-coverage, it's new that we have a github action for that. It would be nice to have a unit test for `Gbasf2Process._build_gbasf2_submit_command()`, but running the method requires instantiating the class and that requires a gbasf2 installation, so this is a bit involved and I think you can ignore it in this case. Though I think I would try adding a test in the future, it should work if we monkey-patch some methods of `Gbasf2Process` so that we can instantiate it with a gbasf2 environment and run the tests with that....

- [ ] add documentation

b2luigi/batch/processes/gbasf2.py
```diff
@@ -410,7 +410,11 @@ def _build_gbasf2_submit_command(self):
 
         gbasf2_input_dataset = get_setting("gbasf2_input_dataset", default=False, task=self.task)
         if gbasf2_input_dataset is not False:
-            gbasf2_command_str += f" -i {gbasf2_input_dataset} "
+            if ".txt" not in gbasf2_input_dataset:
```
meliache, line 413: I would try to be more explicit here with the [string.endswith](https://www.tutorialspoint.com/python/string_endswith.htm) method:
```suggestion
            if not gbasf2_input_dataset.endswith(".txt"):
```
- [ ] replace `in` with `.endswidth`

ArturAkh: Hi @meliache

Yeah, I've actually noticed shortly after my pull request, that you have already done some work in #75 on that issue (see my comment on it).

I'm fine with both ways to solve that & can test the branch in #75 if you like, since I am currently anyhow working on a suitable use-case: https://github.com/ArturAkh/b2luigi/tree/fei_with_gbasf2

In that case, I have a working environment with all the needed software.

Please just decide which solution should be taken to the main branch and I'll either test the #75 one or extend the needed "help" message adaption & the suggested code change.

meliache: About your solution I just don't like that in theory the `--input_dslist` works also with LFN files that have other endings than `.txt`, e.g. I know some people [use[(https://questions.belle2.org/question/8334/is-it-possible-to-process-one-experiment-on-the-grid-by-one-command/) `.lfn`. So someone might provide the path to their local text file with another file ending and wonder why they get an error that the dataset doesn't exist.

About my solution I I don't like that I add yet another setting and the complexity of having an `xor`-like settings requirement. So I need to sleep a night over this and maybe ask someone.

meliache:
Duplicate of #75, continue work there.

PR #94, opened 2021-04-27
Automatic FEI training workflow with b2luigi using gbasf2

ArturAkh:


examples/gbasf2_fei_training/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+    "gbasf2_install_directory": "/work/akhmet/Belle2Project/gbasf2",
```
meliache, line 2: When you finalize the PR it would be better to leave all settings that are specific to you personal setup at the default values. E.g. `gbasf2_install_directory`, `result_dir`, etc. Though it should be clear anyway that this is just an example

examples/gbasf2_fei_training/fei_grid_workflow.py
```diff
@@ -0,0 +1,473 @@
...
+
+        path = create_fei_path(filelist=[], cache=self.cache, monitor=self.monitor)
+
+        if self.stage > -1:
+            # remove symlinks and not needed Summary.pickle files
+            for key in self.get_input_file_names():
+                if key == "mcParticlesCount.root" or key.endswith(".xml"):
+                    adjusted_key = '"' + key + '"'
+                    os.system(f"rm {adjusted_key}")
+            os.system("rm -f Summary.pickle*")
```
meliache, line 186: use `os.remove` (see same suggestion in #91)

examples/gbasf2_fei_training/fei_grid_workflow.py
```diff
@@ -0,0 +1,473 @@
...
+
+
+class ProduceStatisticsTask(luigi.WrapperTask):
+
+    def requires(self):
+
+        # yield FEITrainingTask(
+        #     mode="Training",
+        #     stage=0,
+        # )
```
meliache, line 444: I'm not sure why large chunks of code are commented out but when you finalize this PR, all comments should serve documentation purposes, either documenting a part of code or showing an example for an alternative. The examples are long anyway

examples/gbasf2_fei_training/fei_grid_workflow.py
```diff
@@ -0,0 +1,473 @@
...
+        dataset_sites = [site.strip() for site in open(self.get_input_file_names('dataset_sites.txt')[0], 'r').readlines()]
+        completed_replicas = []
+        for ds_site in dataset_sites:
+            completed_replicas.append(run_with_gbasf2(shlex.split(f"gb2_ds_rep {foldername}/"
+                                      f"sub00 -d {ds_site} -s {self.remote_initial_se} --force")))
+
+        if sum([proc.returncode for proc in completed_replicas + [completed_copy]]) == 0:
+            with open(f"{self.get_output_file_name('successfull_input_upload.txt')}", "w") as timestampfile:
+                timestampfile.write(timestamp)
+                timestampfile.close()
```
meliache, line 434: Closing the file explicitly is not necessary because it is closed automatically when the `with...` scope ends. Actually the `with` is much better because it closes the file even if an exception occurs during writing.
```suggestion
            with open(f"{self.get_output_file_name('successfull_input_upload.txt')}", "w") as timestampfile:
                timestampfile.write(timestamp)
```

examples/gbasf2_fei_training/fei_grid_workflow.py
```diff
@@ -0,0 +1,473 @@
...
+        taroutname = self.get_output_file_name("fei_analysis_inputs.tar.gz")
+        taroutdir = os.path.dirname(taroutname)
+        outputsstring = ' '.join(['"'+o+'"' for o in outputs])
+        tarcmd = f"rm -f {self.get_output_file_name('successfull_input_upload.txt')}; "
+        tarcmd += f"cp {outputsstring} {taroutdir}; "
+        tarcmd += f"pushd {taroutdir}; "
+        baseoutputsstring = ' '.join(['"'+os.path.basename(o)+'"' for o in outputs])
+        tarcmd += f"tar -vczf {taroutname} {baseoutputsstring}; "
+        tarcmd += f"rm -f {baseoutputsstring}; popd"
+        os.system(tarcmd)
```
meliache, line 416: This bash programming from python is difficult to read for me, even if it works when you first hack together.  I would replace the calls to shell command with proper python. Why use bash when we have the power of python? Use the built-in [tarfile](https://docs.python.org/3.7/library/tarfile.html) package for unpacking, `shutil.copy` for copying and `os.remove` for removing etc. Using python ensures that the code runs for users with non-bash shells and on other operating systems (though officially we only support linux anyway). Also it ensures proper error handling in case one of the command fails, and the user will see which python function exactly failed. Maybe if I take the time to understand what exactly the bash commands are doing I will replace it myself.

Anyway, since this is just an example and not part of the b2luigi codebase I'm not that strict here. At the very least I would replace `os.system` with the newer `subprocess.call(shell=True)` (I think your commands with `pushd`, `popd` etc. demand a shell, if you don't change them).

ArturAkh:
@meliache

Thanks a lot for your review! I'll just very used to do `os.system` when I would like to do something bash-like :D

I'll try to introduce your suggestions to the example in the following days.

examples/gbasf2_fei_training/fei_grid_workflow.py
```diff
@@ -0,0 +1,473 @@
...
+            additional_file = os.path.join(luigi.get_setting("remote_tmp_directory").rstrip('/')+timestamp,
+                                           "stage"+str(self.stage - 1), "sub00", "fei_analysis_inputs.tar.gz")
+            luigi.set_setting("gbasf2_input_datafiles", [additional_file])
+
+            # create symlinks to files, which are needed for current FEI analysis stage
+            for key in self.get_input_file_names():
+                if key.endswith(".xml"):
+                    filepath = '"' + self.get_input_file_names(key)[0] + '"'
+                    adjusted_key = '"' + key + '"'
+                    os.system(f"ln -sf {filepath} {adjusted_key}")
```
meliache, line 174: Replace with `os.symlink` and if you want to force it, you can use the following function, which I found by googling [here](https://mail.python.org/pipermail/python-list/2000-March/055125.html) and adapted it to python3:

``` python
import errno

def force_symlink(file1, file2):
    try:
        os.symlink(file1, file2)
    except OSError as e:
        if e.errno == errno.EEXIST:
            os.remove(file2)
            os.symlink(file1, file2)
    ```

File: examples/gbasf2_fei_training/fei_grid_workflow.py [hunk 7]
```diff
@@ -0,0 +1,473 @@
...
+from b2luigi.basf2_helper.tasks import Basf2PathTask
+from b2luigi.basf2_helper.utils import get_basf2_git_hash
+from b2luigi.batch.processes.gbasf2 import run_with_gbasf2
+
+from B_generic_train import create_fei_path, get_particles
+import fei
+
+
+def shell_command(cmd):
+    os.system(cmd)
```
meliache, line 20: Why ??? I don't see where this is used and I don't see what this adds besides aliasing a short command with a slightly longer command. Anyway, `os.system` is outdated and not recommended. If you _really_ think you need to run your command in a shell, use `subprocess.run(cmd, shell=True)` (`subprocess.call(cmd, shell=True)` also works, `run` is newer and the currently recommended subprocess function). But if you command doesn't need to run in a dedicated shell (bash/zsh/...), try running it without `shell=True`, e.g. just
``` python
cmd_list = shlex.split(cmd_str)
subprocess.run(comd_list, check=True)
```

meliache:
Thanks for the PR. I made some early comments about the code-style, mostly they concern using different `os.system` calls instead of using the python functions for that. In principle we are less strict with the examples as they are not part of b2luigi, this is why the tests don't cover them. Anyway, since we use the examples as advertisements, I would suggest trying for clean, idiomatic python code (also maybe PEP8-ify them with a formatter like `black`, `yapf` or `autopep8`).

Once the example looks nice, we can consider whether we want to include it somewhere in the online sphinx documentation under [Belle II Specific Examples](https://b2luigi.readthedocs.io/en/stable/advanced/basf2-examples.html)

meliache:
As I see it, this PR is not urgent, so I might help with polishing when I have more time.

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: ok, thanks! corrected

examples/gbasf2_fei_training/settings.json
ArturAkh: I would prefer to leave the settings explicit - changed to some dummy paths to be replaced, of cource - since I plan to refer to them in the documentation of this example.

I'll do that in one of the following commits

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: Changed.

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: I'd prefer to keep these example, how the workflow can be started to have them all at hand.

The main one would be the one currently activated, but for testing purposes, the ones in the comments can be used.

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: The aim is a tarball containing the needed input files from different directories **without** their directory structure (just the files), and the tarball should be put into the output directory of the task.

I had a few trials with the `tarfile` package, but I haven't managed to have it done as I needed it to be. The shell command construction was the nearest solution to overcome that.

ArturAkh: I'll try to rethink this a bit to solve it

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: yeah... this was my solution to use it for multiprocessing. I'll redo that in a proper way. And here, I need to perform a shell command ton run the fei merging executable.

meliache: I don't see where the function `shell_command` does multiprocessing, as far as I see it there is no advantage to using `os.system` directly, it's basically an alias that you created.

We already a [Basf2nTupleMergeTask](https://github.com/nils-braun/b2luigi/blob/main/b2luigi/basf2_helper/tasks.py#L115) which interally calls `fei_merge_files`/`analysis-fei-mergefiles` via`subprocess.check_call` (equivalent to the newer `subprocess.run(check=True`)  and this works perfectly well as a mere subprocess without requiring a shell with e.g. `shell=True`.

ArturAkh: The function `shell_command` does no multiprocessing, it is **used** for multiprocessing:

https://github.com/ArturAkh/b2luigi/blob/fei_with_gbasf2/examples/gbasf2_fei_training/fei_grid_workflow.py#L227-L228

Since I have very specific requirements for my merging building block, I'd like to not inherit from `Basf2nTupleMergeTask` if possible.

I would rather consider that as a future improvement to put it better together - after this pull request is merged.

ArturAkh: As said before, I'll the multi-processed merge command in a better way.

ArturAkh:
Concerning this PR not being urgent: I'd like to finish the implementations (and all adjustments for better python style) by the end of this week, since from then on, I'll work on the documentation.

It is very important to make a working version of this example available to the community to run FEI on grid, therefore the time-constraints.

That being said, I'll work on further improvements until the end of the week, such that we are all satisfied with its status, but then I would like to merge this example into the main repo.

Otherwise I would have to refer to my own fork in the documenation of this example.

examples/gbasf2_fei_training/fei_grid_workflow.py
meliache: By using `p.map` doesn require defining your own `shell_command` function, you can just use something like
``` python
 	        p = Pool(self.ncpus)
	        p.map(subprocess.check_all, cmds)
```
Experience python programmers should recognize `subprocess.check_call`.

Only issue is that subprocess.run requires a command as a list instead of as a string, but you can split strings into command lists via `shlex.split(cmd)`.

ArturAkh:
Concerning the place of the example documentation. I think it is better to put it into the `basf2` documenation, as also noted in the example [README.md](https://github.com/ArturAkh/b2luigi/blob/fei_with_gbasf2/examples/gbasf2_fei_training/README.md). Reason: this documentation on FEI is most presumably more often used, than b2luigi example, I suppose.

Of course, a link to the documentation can be put to the [Belle II Specific Examples](https://b2luigi.readthedocs.io/en/stable/advanced/basf2-examples.html)

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: As said before: I will rework the command for sure...

And yes, when it comes to proper python style, I'm not that experienced. I'm already satisfied, when thing work as I expect them to work.

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: Corrected. I really hope, that the python-based command can cope properly with the naming of the FEI training files... The shell command did the way I've constructed them.

examples/gbasf2_fei_training/fei_grid_workflow.py
meliache: :+1: This is a valid reason as it roughly falls into the documentation purposes, as it shows what else could be used.

examples/gbasf2_fei_training/settings.json
meliache: I understand that, so it's to showcase the settings where have available. I admit settings discoverability is bad at the moment.

meliache:
>  think it is better to put it into the `basf2` documentation, as also noted in the example [README.md](https://github.com/ArturAkh/b2luigi/blob/fei_with_gbasf2/examples/gbasf2_fei_training/README.md).
Then I don't understand why not include the whole example in basf2, so that you can include it in the sphinx without having to use a link? The examples here are supposed to be examples how to use b2luigi, not examples how to use the FEI.

ArturAkh:
It is included here, because it is b2luigi based...

In one or the other way, I would like to avoid investing double work into documentation. And I think it is more suitable in the FEI documenation of `basf2`. To be discussed, as soon as I start with the documentation. We are not at that point yet.

examples/gbasf2_fei_training/fei_grid_workflow.py
ArturAkh: Ok, should be resolved now.

examples/gbasf2_fei_training/fei_grid_workflow.py
meliache: Actually maybe it's easier to do with [shutil.make_archive](https://docs.python.org/3.7/library/shutil.html#shutil.make_archive):
``` python
        shutil.make_archive(
            base_name=<base name of output tar file without the file ending>,
            format="gztar",
            root_dir=<directory in which the tar.gz archive should be created>,
            base_dir=<directory to archive>,
            verbose=True,
            dry_run=False,
        )
```
I looked at my quality indicator training script which is in basf2 and I had used this function to make an archive (see [here](https://stash.desy.de/projects/b2/repos/software/browse/tracking/scripts/tracking/train/combined_quality_estimator_teacher.py#2452-2458))

ArturAkh: I actually have now a solution with `tarfile` :D, at least I think it will work out

ArturAkh: Let me put it together with `tarfile`, and then we can decide what's better :)

ArturAkh: Ok I think in that way, it should work out: https://github.com/ArturAkh/b2luigi/blob/fei_with_gbasf2/examples/gbasf2_fei_training/fei_grid_workflow.py#L410-L419

examples/gbasf2_fei_training/settings.json
ArturAkh: Ok created a proper template-like example for settings

meliache:
> It is included here, because it is b2luigi based...

But it is also FEI-based and basf2-based. b2luigi can be installed easily from anywhere via pip. Several people are using b2luigi in gbasf2 for their purposes, e.g.  for their analyses, for the systematic corrections framework, or for the quality indicator training script that I once created. These use-cases are very specific and therefore didn't land here. I would like more examples, but for showing people how to apply b2luigi to different classes of common problems and for the purpose of the online-documentation.

Anyway, I assume that the largest work consists in creating the python files and I hope it will mean no additional work for you if you decide to publish the examples somewhere else. The sphinx documentation would also be quite portable because sphinx is sphinx.

I'm sorry for nitpicking on refactoring-issues before such a large-picture issue. I will keep the PR open for the moment while you are still working on it but please know that I'll discuss this and think whether we really need this _here_ and I might close it in the future.

Anyway, I'm very happy and always looking forward to your contributions to b2luigi and I'd be interested to see you experiences.

Maybe in the future we could document a list of example projects that use b2luigi and then link to the FEI documentation, the systematic corrections framework etc.

ArturAkh:
Yeah, I understand your point. Examples in `b2luigi` was for me a natural entry-point for the example to be published as code. Otherwise - to be able to document it somewhere - I would need another repository. That's also fine with me, but I thought it would be of benifit to have it within b2luigi.

Anyway, I guess I have incorporated your comments so far - thanks for being picky on the styl, only in that way I will learn proper python-style coding. In the end - if everything works as before - it looks much better than my shell hacks ^^

I'll test it once more on grid to check, whether everything works as expected and then we can see what we do with that pull request.

meliache:
As discussed also outside this PR discussion we decided the example should live outside b2luigi if it's not meant for being included in the b2luigi documentation

ArturAkh:
Fine with me & thanks for review :)

PR #95, opened 2021-04-28
Fix b2luigi __version__ being shadowed by luigi

meliache:
The import statement
``` python
from luigi import *
```
also imports the `__version__` from luigi and this seems to shadow the b2luigi version defined at the top of the file.

The quick fix was to move the `__version__` definition down, but the best would be to no import the luigi version to begin with.

I found this when I checked `b2luigi.__version__` on a remote host to check the installed version and saw 3.0.2 as a result, which is the luigi version. I wonder why I didn't see this earlier...

### DIFF
--- a/b2luigi/__init__.py
+++ b/b2luigi/__init__.py
@@ -2,7 +2,8 @@
 from luigi import *
 from luigi.util import inherits, copies
 
-# version mast be defined importing luigi namespace, otherwise b2luigi.__version__ gives luigi version
+# version must be defined after importing the luigi namespace, 
+# otherwise the b2luigi.__version__ gets overwritten by the one from luigi
 __version__ = "0.6.3"
 
 from b2luigi.core.parameter import wrap_parameter, BoolParameter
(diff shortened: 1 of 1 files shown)


b2luigi/__init__.py
```diff
@@ -1,9 +1,10 @@
 """Task scheduling and batch running for basf2 jobs made simple"""
-__version__ = "0.6.3"
-
 from luigi import *
 from luigi.util import inherits, copies
 
+# version mast be defined importing luigi namespace, otherwise b2luigi.__version__ gives luigi version
```
nils-braun, line 5: Just some minor comment suggestions
```suggestion
# version must be defined after importing the luigi namespace, 
# otherwise the b2luigi.__version__ gets overwritten by the one from luigi
```

nils-braun (approved):
Nice catch!
Completely correct. I just have some minor wording suggestions

meliache:
Upps thanks for checking the comment, I think it shows I was tired and in a hurry when writing this :sweat_smile:

PR #99, opened 2021-04-29
b2luigi/batch/processes/gbasf2.py: add additional required split

ArturAkh:
An addtional split of stdout is required for the case, when there are skipped files for download.

This is corrected now & a corresponding test is added.


meliache:
Looks good to me, thanks a lot for adding the test, actually it serves as a good documentation because by looking at the test output I see what you're trying to achieve with the changed code.

PR #106, opened 2021-05-03
Added 'inherits_without' decorator to enable inheritance of everything except chosen parameters from task

sognetic:
Pretty much what it says on the tin. Maybe this could also replace the existing decorator, however I didn't want to introduce to many changes from stock luigi.

This feature is useful if you want to keep the same order of parameters (so the automatically-generated folders match between tasks) but maybe drop one or two parameters since you have resolved them (e.g. by merging files).

This is more or less a proposal, not necessarily a finished PR.

### DIFF
--- a/b2luigi/__init__.py
+++ b/b2luigi/__init__.py
@@ -7,7 +7,7 @@ from luigi.util import inherits, copies
 __version__ = "0.6.6"
 
 from b2luigi.core.parameter import wrap_parameter, BoolParameter
-from typing import Union, Optional, Union
+from typing import Optional, Union, List
 
 wrap_parameter()
 
@@ -73,7 +73,6 @@ class requires(object):
 
 class inherits_without(object):
 
-
     def __init__(self, *tasks_to_inherit, without: Optional[Union[List, str]] = None):
         super(inherits_without, self).__init__()
         if not tasks_to_inherit:
@@ -85,7 +84,7 @@ class inherits_without(object):
         elif without is None:
             self.without = []
         else:
-            self.without =  without
+            self.without = without
 
     def __call__(self, task_that_inherits):
         # Get all parameter objects from each of the underlying tasks
--- a/b2luigi/__init__.py
+++ b/b2luigi/__init__.py
@@ -74,7 +74,7 @@ class requires(object):
 class inherits(object):
 
     """
-    Similar to `b2luigi.requires`, this copies the luigi.inherits functionality but allows specifying parameters you
+    This copies the luigi.inherits functionality but allows specifying parameters you
     don't want to inherit.
 
     It can e.g. be used in tasks that merge the output of the tasks they require. These merger tasks don't need
@@ -107,6 +107,8 @@ class inherits(object):
 
     Parameters:
         without: Either a string or a collection of strings
+
+    See also: `b2luigi.requires` which extends ``luigi.requires``.
 
     """
 
--- a/b2luigi/__init__.py
+++ b/b2luigi/__init__.py
@@ -138,7 +138,7 @@ class inherits(object):
                     self.without.remove(param_name)
 
         # Make sure we're only removing parameters that exist
-        assert len(self.without) == 0, f"You're trying to remove parameter(s) {self.without} which do(es) not exist."
+        assert not self.without, f"You're trying to remove parameter(s) {self.without} which do(es) not exist."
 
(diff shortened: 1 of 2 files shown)


pakozm:
I really like this feature :+1:

nils-braun:
@meliache will you have a look or shall I do the review?

meliache:
@nils-braun Would be glad about your opinion, feel free to do the review. In the following some general thoughts that I had already shared with @sognetic in private:

As I see it, this could equally well be a PR for upstream luigi, there's nothing b2luigi-specific about this. But there's nothing bad about merging here first and later trying to port them to luigi.

To be honest, I'm surprised why this feature or something similar doesn't exist yet in luigi. I was wondering whether there might be a way two achieve something similar with existing functions which we might have just overlooked, but I haven't found anything yet.

Imo this PR can be discussed in terms of user interface and experience design, that is, how we want users to build their workflows.This is why I'm happy to hear other people's opinion.

nils-braun:
Hi @sognetic , @meliache 
I am in general quite happy with the feature. Thank you very much. I have also not found a (simple) way to achieve this with Luigi's own methods, so I like the solution.
I do have a few comments on the implementation etc, but if I understood @sognetic correctly, this PR is only a draft anyway. So I would suggest to go forward with the implementation, add documentation and tests and then I can have another look - what do you think?

Concerning the user experience I definitely see the need for such use-cases. There is only a very small caveat: if I imagine having a task that requires a few other subtask (with different values for parameter a), and this parent tasks merges those files it makes a lot of sense to omit the parameter a from the merge task. However, the information which values of "a" have been computed and merged together is lost then. If one adds another possible a value later, there is no way to see this. However this does not speak against this PR I'm general, as there are multiple ways how to solve this and there will not be a general solution anyways. Maybe however, it would be worth mentioning this in the documentation.

meliache:
@nils-braun @sognetic :+1: for documentation, i.e. class/function docstrings and maybe including an example or changing an existing example to use this decorator, e.g. a simple merger task. I'd also add a unit test, probably it would make sense just extending [test_requires.py](https://github.com/nils-braun/b2luigi/blob/main/tests/core/test_requires.py), which seems rather straight-forward. If you (@sognetic) require help with that, I can also do that or help with that.

sognetic:
@nils-braun @meliache  Ok sure, I'll add some documentation and a test. Thanks for the pointers!

b2luigi/__init__.py
```diff
@@ -68,3 +69,85 @@ def requires(_self):
...
+            for param_name, param_obj in task_to_inherit.get_params():
+                # Check if the parameter exists in the inheriting task and isn't in without
+                if not hasattr(task_that_inherits, param_name) and param_name not in self.without:
+                    # If not, add it to the inheriting task
+                    setattr(task_that_inherits, param_name, param_obj)
+                elif param_name in self.without:
+                    self.without.remove(param_name)
+
+        # Make sure we're only removing parameters that exist
+        assert len(self.without) == 0, f"You're trying to remove parameter(s) {self.without} which do(es) not exist."
```
nils-braun, line 139: More pythonic would be 

```suggestion
        assert not self.without, f"You're trying to remove parameter(s) {self.without} which do(es) not exist."
```

b2luigi/__init__.py
```diff
@@ -68,3 +69,85 @@ def requires(_self):
         task_that_requires.requires = requires
 
         return task_that_requires
+
+
+class inherits(object):
+
+    """
+    Similar to `b2luigi.requires`, this copies the luigi.inherits functionality but allows specifying parameters you
```
nils-braun, line 77: I understand your idea here, but not everyone reads the documentation from top to bottom and the reference to requires might even confuse people (because of course inherits does not do the same than requires). I think you can remove it here and maybe create a "See also" section (or alike) further down.

nils-braun (approved):
Great, @sognetic!
I only have two minor comments.

b2luigi/__init__.py
sognetic: Ok sure, that's definitely more pythonic.

b2luigi/__init__.py
sognetic: Ok, done. However I noticed that the documentation from `b2luigi.inherits` doesn't show up in sphinx and neither does the documentation I added. Where should this show up do you think?

PR #109, opened 2021-06-11
Update luigi scheduler progressbar for gbasf2 tasks

meliache:
Since we now have luigi 3 as a dependency, we can make use of the progressbar-features for the scheduler. Here a small screenshot. The uppermost worker is a local merger task where I set the progressbar manually in its `run` method. The other progressbars are of gbasf2 tasks, each of which supervises a gbasf2 grid project consisting of many individual jobs. So for those, the progressbar that I implemented with this PR indicates how many of those jobs in that project are done. 
grafik

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -283,6 +283,12 @@ class Gbasf2Process(BatchProcess):
         n_failed = n_jobs_by_status["Failed"]
         n_in_final_state = n_done + n_failed
 
+        # try updating progressbar for central scheduler web view
+        try:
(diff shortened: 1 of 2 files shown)


meliache:
Now also added the gbasf2 job status to the central scheduler view.
grafik

nils-braun:
Thanks @meliache - looks nice.
Is there a way to have less digits in the percentage in the progress bar (I did not check the code so far)

meliache:
The function takes a numerical percentage as an input, I haven't checked if it also takes a format string (writing from mobile now), but I could just round the number to some decimal places.

Issue #112, opened 2021-06-13
Failed gbasf2 grid download because failed_files.txt contains directory


meliache:
My task was all successful and tried to download the job outputs, but it failed with the following message:
```
Downloading remaining files from dataset with command  gb2_ds_get --force /belle/user/meliache/2021-06-106ac9f82805/sub00/upsilon_4s_*.root --input_dslist /nfs/dust/belle2/user/meliache/git_hash=light-2002-ichep/
job_name=2021-06-10/input_files=hashed_eb2e36a81e191f2c9fe4b2962bb9de29/mc=True/fei_charge_type=B0/upsilon_4s.root.partial/failed_files.txt
No file found

/afs/desy.de/user/m/meliache/.local/lib/python3.6/site-packages/b2luigi/batch/processes/gbasf2.py:325: RuntimeWarning: RuntimeError('No output data for gbasf2 project 2021-06-106ac9f82805 found.',)
  warnings.warn(repr(err), RuntimeWarning)
```
I checked the output of the `failed_files.txt` and it only contains the line
```
/belle/user/meliache/2021-06-106ac9f82805/sub00/upsilon_4s_00039_job194473040_00.root in /nfs/dust/belle2/user/meliache/git_hash=light-2002-ichep/job_name=2021-06-10/input_files=hashed_eb2e36a81e191f2c9fe4b2962bb9de29/mc=True/fei_charge_type=B0/upsilon_4s.root.partial/2021-06-106ac9f82805/sub00
```
Seems to be caused by PR #91 by @ArturAkh, which introduced this log file and included some command output parsing, which might have failed. As I assume from looking at the helper method `_failed_files_from_dataset_download` it should only contain root files, but maybe something went wrong in all the string splitting?

ArturAkh:
Hi @meliache,

Yes, it may well be that this is because of parsing, since this seems to be a wrong line taken into the text file.

Do you use the newest `gbafs2` release by now? (which may have changed the printout format from v5.0.2 to v5.1.0)

In case you still have the printout in your screen/tmux session at the point where the first download attempt was started (so the one leading to creation of the file `failed_files.txt`), it would be good if you could send it to me.

For the time-being, I'd suggest to remove the file & restart the download again (reproducing previous behaviour of trying to download all output files again).

meliache:
[...]

PR #113, opened 2021-07-01
Fixing parsing problems for failed downloads due to new syntax

ArturAkh:
Fixes issue #112

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -626,11 +626,6 @@ class Gbasf2Process(BatchProcess):
             split("Skip ")[0].strip().split('\n')
         failed_files = [line.split('.root')[0] + '.root' for line in failed_files
                         if len(line.strip()) > 0 and len(line.split('.root')) > 1]
-
-        # Make consistency check, that all lines contain paths to ROOT files
-        for entry in failed_files:
-            if not entry.endswith('.root'):
-                return None
         return failed_files
 
     def _download_dataset(self):
@@ -686,14 +681,6 @@ class Gbasf2Process(BatchProcess):
                 with open(monitoring_failed_downloads_file, 'w') as ffs:
                     ffs.write("\n".join(failed_files))
             else:
-                if failed_files is None:
-                    stdout_file = os.path.join(tmp_output_dir_path, "download_stdout.txt")
-                    print("WARNING: Failed to create a valid .txt file with ROOT files from failed downloads as entries.")
-                    print(f"Passing all stdout from the download command to {stdout_file}.")
-                    print("Please send it to b2luigi developers for debugging.")
-                    print("After restarting, the workflow will proceed with default download behaviour.")
-                    with open(stdout_file, 'w') as sof:
-                        sof.write(stdout)
                 try:
                     os.remove(monitoring_failed_downloads_file)
                 except OSError as e:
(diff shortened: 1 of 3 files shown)


meliache:
Thanks a lot. I like that you didn't remove the old test outputs, so I know that this is still backwards compatible to the old releases. :+1: Not sure if this will work for future releases, maybe we can be a bit smarter than splitting on `" to "`,  but anyways it's an improvement on the previous version and fixes a bug, so I think I can merge this as is.

b2luigi/batch/processes/gbasf2.py
```diff
@@ -624,7 +624,7 @@ def _failed_files_from_dataset_download(self, stdout):
         failed_files = stdout.split('Failed files:')[-1].\
             split("Files with duplicated jobID, not downloaded:")[0].\
             split("Skip ")[0].strip().split('\n')
-        failed_files = [line for line in failed_files if len(line.strip()) > 0]
+        failed_files = [line.split(' to ')[0].strip() + '.root' for line in failed_files if len(line.strip()) > 0]
```
meliache, line 627: 1. Instead of `if len(line.strip()) > 0`, maybe we should check that the new substring before the `' to '` is not empty, that is 
``` python
if len(line.split(' to ')[0].strip()) > 0
```
2. I just noticed that probably we can use the implicity booleaness of strings for a more pythonic check, together with the suggestion from 1. that would be 
``` python
if line.split(' to ')[0].strip()
```
3. I can imagine that in a future release the `' to '` output might change to another string (e.g. to an arrow), but I expect that the file names will always be root files. Therefore, instead of splitting on ` to ` and appending `.root`, I think it would be more future safe to select the text in each line upto including `.root`.  For that on can either use a regex with the `re` package or just split on `.root` and adding it again.

Maybe I'll add some suggestions tomorrow that you can just accept.

ArturAkh: @meliache I'll add the suggestions 1 and 2 directly, I think doing it in what way looks better.

Concerning suggestion 3: yeah this seems still to be bugged the adding of `.root` should not be there in case of using ` to ` for splitting. The reason why it is still there is because I had splitting by `.root` in the commits before but thought, that might be too specific for output file extension. However, if you think that covers most of the application cases, I will introduce the splitting by `.root` back

ArturAkh: @meliache  I've had an example with a failed file. Seems like there is a bug on `gb2_ds_get` side, that the option `--input_dslist` corrupt and `No file found` is printed out. I'll have a look at that.

ArturAkh: So all in all, my fix won't work although now better catching the case, but it is a problem on `gbasf2` side. I'll have a look at the code and will report here the fix.

ArturAkh: Found the issue...

https://stash.desy.de/projects/B2DC/repos/belledirac/browse/Client/controllers/datasetCLController.py#142

`line[:-1]` should be replaced with `line.strip()`...

I've created a JIRA Ticket for that https://agira.desy.de/projects/BIIDCD/issues/BIIDCD-1408

ArturAkh:
Any further actions for this PR required?

meliache:
> Any further actions for this PR required?
Nope, LGTM, for some reason I thought you maybe want to add something or maybe I also just forgot about this PR, thanks for the reminder.

ArturAkh:
> > Any further actions for this PR required?
> 
> Nope, LGTM, for some reason I thought you maybe want to add something or maybe I also just forgot about this PR, thanks for the reminder.

Ok, thanks for the merging in the PR!

Issue #117, opened 2021-07-26
Rename `MasterTask`'s in examples to something less culturally sensitive


meliache:
We have already made our main branch `main` and we could follow through by replacing all examples in the documentation where the implementation of `luigi.WrapperTask`  is named `MasterTask` by something else. Suggestions:

- `MainTask`
- `WrapperTask` -- In this case the base name would be the same as the class it inherits from, just in a different namespace. Maybe it would be better a meaningful prefix to convey what it does, e.g. `ReconstructionWrapperTask` or `AnalysisWrapperTask`

philiptgrace:
The systematic corrections framework uses `AggregatorTask`, which I quite like.

nils-braun:
Both are fine with me - thanks for this reasonable change @meliache

PR #118, opened 2021-07-26
Give instances of `MasterTask` in doc examples less sensitive names

meliache:
I already mentioned the rationale in #117 . I think it would be more appropriate to use other task names than `MasterTask`, which doesn't convey much meaning anyway. In one instance I named it `MainTask`, similar to how we renamed the main branch name in #39. But since we are implementing `luigi.WrapperTask`, it might be more meaningful to encode what is being wrapped in the name, e.g. `AnalysisWrapperTask`, which I used in the other two occurances.

This is just a documentation change, but I think it is important because the examples are how we present b2luigi, it's the first of b2luigi that new users see and often they just adopt the naming from the examples.

It's not urgent, so can wait for a comment from @nils-braun who did the branch renaming in #39, maybe you have some ideas with regards to proper naming.

### DIFF
--- a/docs/advanced/basf2-examples.rst
+++ b/docs/advanced/basf2-examples.rst
@@ -141,14 +141,14 @@ nTuple Generation
             return path
 
 
-    class MasterTask(luigi.WrapperTask):
+    class AnalysisWrapperTask(luigi.WrapperTask):
         def requires(self):
             # somehow loop over the runs, experiment etc.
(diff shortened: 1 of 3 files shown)


nils-braun (approved):
Thanks @meliache 
The chosen names are already good, also AggregationTask is a good suggestion. our van choose, I am fine with both

PR #121, opened 2021-09-02
Replace parsing of gb2 command output with DIRAC API calls

philiptgrace:
Addresses tasks in #47.
* Replaced `check_if_dirac_proxy_is_initialized.py` with more general `gbasf2_proxy_info.py` that can be used to get the DIRAC username and the remaining proxy lifetime.
* Wrote `gbasf2_ds_list.py`, which can be used in place of `gb2_ds_list`. This script prints a JSON-encoded list, so manual parsing of `gb2_ds_list` is no longer required.

I have run tests myself that these behave as expected.

Outstanding task:
* `_failed_files_from_dataset_download` still parses the output of `gb2_ds_get`. No idea how difficult it would be to replace this with a script similar to the ones in this PR. Annoyingly, a lot of the BelleDIRAC framework is written in functions that return strings to be printed, rather than returning the result in a useful data structure.


meliache:
Thanks for this PR. You seem to be definitly more familiar with `BelleDirac` than me.. Sorry for being a bit slow, I'm currently on holiday until including the 7th of September. On a first glance I didn't see anything to critisize, you code looks very clear and good to me. If it's not urgent I'll put it on my todo list and give this a proper look after I'm back from holiday and more sober than now. If forgot reviewing this PR, just message me in the Belle II chat.

Cheers, Michael

meliache:
This PR still looks very good after my holiday. Thanks a lot!

Issue #48, opened 2020-10-09
Improve dirac proxy validity time handling


meliache:
When working with gbasf2 subprocesses in the gbasf2 BatchProcess, in addition to a gbasf2 environment most commands also need an alive dirac proxy. On the command line, users initialize it with `gb2_proxy_init -g belle`, then they get asked for their certificate pasdsword and then it stays alive for 24 hours. So in my wrapper, I check if there is still an alive/valid dirac proxy on the system and run the initalize command only if there is not. The main reason I did that was to ensure the proxy is valid but also I did 
want to reduce superfluous password prompts.

@philiptgrace made me aware [here](https://github.com/nils-braun/b2luigi/pull/46#issuecomment-705289731) that users can set the proxy validity time explicitly via `-v HH:MM` and this can surprisingly be much more than 24 hours. He also suggested [here](https://github.com/nils-braun/b2luigi/pull/46#issuecomment-705915640) that we could consider renewing the proxy not when the remaining validity time reaches 0, but before that for increased stability.

> In addition to having a user option called something like gb2_proxy_time (to be passed to the -v flag), it may be useful to have an option gb2_proxy_minimum_time (defaulting to 0), to re-initialise the proxy if the time left is less than the given number of hours. The usecase I am imagining is a long submission/monitoring/downloading time where the user doesn't want their proxy to timeout part way through.

Though I currently have a branch in progress where I want to implement that this check is done for every `run_with_gbasf` call, so I'm not sure whether this is important, but still I don't see why not. 

- [ ] add setting for proxy validity time
- [ ] maybe increase default time
- [ ] re-initialize dirac proxy before remaining validity time reaches 0 and add setting for that minimum time

philiptgrace:
I cherry-picked your commits and implemented the default & minimum proxy lifetimes on my branch [feature/use-gbasf2-subprocess-to-check-if-dirac-proxy-is-alive](https://github.com/philiptgrace/b2luigi/tree/feature/use-gbasf2-subprocess-to-check-if-dirac-proxy-is-alive).

PR #127, opened 2021-09-16
Improve dirac proxy validity time handling

philiptgrace:
Added two new settings, both used in `setup_dirac_proxy`.
* `gbasf2_proxy_lifetime`: Defaults to 24. When initializing a proxy, set the lifetime to this number of hours.
* `gbasf2_min_proxy_lifetime`: Defaults to 0. During processing, prompt user to reinitialize proxy if remaining proxy lifetime drops below this number of hours.

Addresses #48

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -954,7 +954,10 @@ def setup_dirac_proxy():
     if get_proxy_info().get("secondsLeft", 0) > 3600 * get_setting("gbasf2_min_proxy_lifetime", default=0):
         return
     # initialize proxy
-    hours = int(get_setting("gbasf2_proxy_lifetime", default=24))
+    lifetime = get_setting("gbasf2_proxy_lifetime", default=24)
+    if not isinstance(lifetime, int) or lifetime <= 0:
+        warnings.warn("Setting 'gbasf2_proxy_lifetime' should be a positive integer.", RuntimeWarning)
+    hours = int(lifetime)
     run_with_gbasf2(shlex.split(f"gb2_proxy_init -g belle -v {hours}:00"), ensure_proxy_initialized=False)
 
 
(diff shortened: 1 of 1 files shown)


b2luigi/batch/processes/gbasf2.py
```diff
@@ -946,10 +951,11 @@ def setup_dirac_proxy():
     Runs ``gb2_proxy_init -g belle`` if there's no active dirac proxy. If there is, do nothing.
     """
     # first run script to check if proxy is already alive or needs to be initalized
-    if get_proxy_info().get("secondsLeft", 0) > 0:
+    if get_proxy_info().get("secondsLeft", 0) > 3600 * get_setting("gbasf2_min_proxy_lifetime", default=0):
```
meliache, line 954: This got me thinking, we could allow for floating point lifetimes and extract minutes as well, e.g.
```suggestion
    lifetime_setting = get_setting("gbasf2_proxy_lifetime", default=24)
    hours = int(lifetime_setting)
    minutes = int(float(lifetime_setting) % 1 * 60)
```

And later set
``` python
run_with_gbasf2(shlex.split(f"gb2_proxy_init -g belle -v {hours}:{minutes}"), ensure_proxy_initialized=False)
```
If we do so, that should also be documented.  But not sure if it's worth adding the complexity for that. rarely anyone needs minute resolution for the proxy lifetime. However, if we don't do that, I would suggest a warning or exception of the lifetime is not integer.

meliache:
Thanks a lot for all the help with the gbasf2 wrapper in b2luigi helping with the issues that I never got to. I really feel know that the gbasf2 part finally has a bus factor > 1. This PR looks good to me, I have a minor suggestion (see comment above), you can take a look and make a comment if you think it's a good idea or not (I'm not sure). Then, I will be happy to merge this :)

b2luigi/batch/processes/gbasf2.py
philiptgrace: I don't really see much use of having minute-level resolution. The use case I have in mind is setting the lifetime for a few days so it doesn't timeout during monitoring/downloading.

A warning for non-integer input makes sense to me.

PR #128, opened 2021-09-20
Test `ignore_additional_command_line_args` option

meliache:
Tests for the ` `ignore_additional_command_line_args` option to `b2luigi.process` introduced in #55 by @klieret. Then I will have more peace of mind when people claim to have issues with this option.

If `ignore_additional_command_line_args=True`, `b2luigi.process` should allow additional command line args not known to b2luigi, as they might be used by the user script. Otherwise, only b2luigi cli args should be allowed.

In this PR there is a bit of boilerplate code due to `DoNothingTask` being defined twice in two test-scripts. Somehow I had issues with using relative import in the test-directory since it's not a package. Originally I had them in a `cli_tasks.py` file in the same directory and imported them with
``` python
from .cli_tasks import DoNothingTask
```
But this didn't work when running the unit tests. Not sure why that is not working but `import ..helpers` works, maybe I should add an `__init__.py` or something like that?  Maybe @nils-braun or someone else knows?

### DIFF
--- a/tests/cli/process_with_user_args_ignored_by_b2luigi.py
+++ b/tests/cli/process_with_user_args_ignored_by_b2luigi.py
@@ -1,0 +1,40 @@
+import argparse
+
+import b2luigi
+
+
+class AlwaysExistingTarget(b2luigi.Target):
+    """
+    Target that always exists
+    """
+
+    def exists(self):
+        return True
+
+
+class DoNothingTask(b2luigi.ExternalTask):
+    """
+    Dummy task that is always already complete and needs nothing to be done.
+
+    Useful for testing ``b2luigi.process`` just for its CLI argument parsing
+    with some dummy task.
+    """
+
+    def output(self):
+        return AlwaysExistingTarget()
+
+
+if __name__ == "__main__":
+    # add some dummy user args. They should work with the ignore_additional_command_line_args option
+    parser = argparse.ArgumentParser("")
+    parser.add_argument("--hello", help="Keyword argument, should be ``'world'``")
+    parser.add_argument("is_one", help="First positional argument, should be ``1``")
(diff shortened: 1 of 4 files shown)


klieret:
Maybe you should force everyone who does a feature PR to also write a test for it :)

meliache:
@klieret Regarding relative imports, I usually never use them in python package code, but the `tests` directory isn't a module of the `b2luigi` package, so I can't import it. `absolufy-imports` converts my relative import statement (not working) to
``` python
from tests.cli.cli_tasks import DoNothingTask
```
That doesn't make sense, I would expect rather something like `from b2luigi.tests.cli.cli_tasks
 import ...`, but both of those result in `ModuleNotFoundError`'s.
 
 Still, in general `absolufy-imports` a nice package that could be added to our `pre-commit` definitions, since self-fixing hooks rock :)

klieret:
Ah, speaking of hooks: I'm surprised you're not using the CI for pre-commit to check them in PRs, as it's as easy as 3 mouse clicks: http://pre-commit.ci/

meliache:
@klieret We _are_ using `pre-commit` hooks to check PR's but it's a task in the `test.yml` github action and not a separate action. However I didn't do this via 3 clicks but by manually adding the task to the yaml, so I will take a look at the `pre-commit-ci` app and see if we can do that in a "better" way. Maybe also with a separate shield in the readme to make it more obvious that we use pre-commit.

The only downside is that sometimes I get PR's that follow with multiple "fix style" commits because instead of installing pre-commit locally, first-time contributers often push commits until pre-commit checks pass on github, despite the instructions in the [development docs](https://b2luigi.readthedocs.io/en/stable/advanced/development.html#local-development).

PR #136, opened 2021-10-12
Remove unused `Gbasf2Process` helper method to capture failed files from stdout

meliache:
This method is not needed since #132  by @ArturAkh, so we can remove it. While at it I did two small refactoring changes to reduce the number of pylint warnings.


ArturAkh:
@meliache sorry for that one, I've forgotten to remove :(

PR #150, opened 2021-11-02
Gbasf2: Fix moving of downloaded multi-sub datasets

### LINKED ISSUE #147: Issue with large projects separated into subfolders
Hello,

thanks for adding the support for multiple sub-directories in the grid projects, i.e. that have multiple `sub*` folders in their output, but unfortunately, I think there is a bug linked to it:
When `b2luigi` moves the downloaded output from temporary folders of `sub00`, `sub01`, `subXX`,  to the output folder, let's call it `result/`, the contents of `sub00` folder is not in the `result/`, but I suspect that it is in the `result/sub00` folder.
I tried to run the  [this code](https://github.com/nils-braun/b2luigi/blob/a81a814821b3380c5504ccd95fb613f3ad5a34c3/b2luigi/batch/processes/gbasf2.py#L711) in the python interpreter using `sub00` and `sub01` input folders and not existing output folder `./result` and I get the following tree:
```
./result
    sub01_content.root
    sub00/
         sub00_content.root
```
This is not expected by the tasks downstream and I would prefer that the output folder would contain all files from all `sub*` folders directly. There shouldn't be any conflicts in the file names, but one has to check for that anyway.

meliache:
For datasets with multiple subs, when moving the contents of the temporary project download directory to the final ouptut dir, previously, there was an error which led to the file structure:

    <result_dir>/B.root/job_name*B.root  # sub00 contents
    <result_dir>/B.root/sub01/job_name*B.root
    <result_dir>/B.root/sub02/job_name*B.root

This commit should fix this back to

    <result_dir>/B.root/job_name*B.root  # contents of all subs

Further, since this moving function turns out to aggregate many lines of code, I moved it into an own helper function, which can be independently tested.

Draft so that I can link to it, but I haven't finished yet writing unit tests.
- [x] Add unit-tests for multi-sub directory structures with dummy root files
- [x] Run a job on the  grid with one input file to check that the single-sub case still works.

Resolves #147

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -656,7 +656,7 @@ class Gbasf2Process(BatchProcess):
             os.makedirs(tmp_output_dir_path, exist_ok=True)
 
             # Need a set files to repeat download for FAILED ones only
-            monitoring_failed_downloads_file = os.path.join(tmp_output_dir_path, "failed_files.txt")
+            monitoring_failed_downloads_file = os.path.abspath(os.path.join(tmp_output_dir_path, "failed_files.txt"))
             monitoring_download_file_stem, monitoring_downloads_file_ext = os.path.splitext(monitoring_failed_downloads_file)
             old_monitoring_failed_downloads_file = f"{monitoring_download_file_stem}_old{monitoring_downloads_file_ext}"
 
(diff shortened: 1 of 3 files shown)


meliache:
Added some unittests on dummy directory strucutres and now this should be ready. The only problem I have currently is that when I run test-jobs on the grid I get download failures, but which according to the mailinglist seem to be caused by problems on the SE sites, so I can't test at the moment.

PR #162, opened 2022-01-08
Use setting for gbasf2 setup file instead of install directory

meliache:
Usage of `gbasf2_install_directory` will cause a pending deprecation warning,
but it can still be used and then the corresponding path to
the `setup.sh` will be derived.

## Rationale:

b2luigi only needs to know which file to source to get the correct
environment. The location or name of that file withing a gbasf2 install
directory can change between gbasf2 releases. Providing the file path directly
results in more simplicity and flexibility.

## Additional enhancement:

In addition, the setting is now only obtained in the `Gbasf2Process` class and
not in the free helper functions, which now get the value via function
arguments. With this, the setting can be provided via a Task class attribute and
the value from the Task attribute will be used everywhere as expected.
This resulted in much more boiler plate code, but also it made my helper
function much more usable as library function without b2luigi settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,15 @@ Older entries have been generated from github releases.
 New entries aim to adhere to the format proposed by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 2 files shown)


meliache:
Partly addresses #195 in that this helps setting the setupfile directly. But it seems there are remaining issues with the current gbasf2 releases that are not fixed by this and I don't have time to fix that right now.

Issue #164, opened 2022-01-24
Bug with htcondor job status property


eckerpatrick:
In #158 the `get_job_status()` function in the subclass `HTCondorProcess` was changed to a property and is therefore no longer callable.
Nevertheless it gets called in https://github.com/nils-braun/b2luigi/blob/main/b2luigi/batch/processes/__init__.py line 135.
I think this is a bug.

meliache:
Thanks to both of for finding the source of the error and for fixing it so fast :+1:

PR #165, opened 2022-01-24
Set get_job_status as function

mschnepf:
Fix issue https://github.com/nils-braun/b2luigi/issues/164

### DIFF
--- a/b2luigi/batch/processes/htcondor.py
+++ b/b2luigi/batch/processes/htcondor.py
@@ -151,7 +151,6 @@ class HTCondorProcess(BatchProcess):
 
         self._batch_job_id = None
 
-    @property
     def get_job_status(self):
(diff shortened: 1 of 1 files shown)


meliache:
LGTM, thanks. Sorry for missing this in the original PR in review. Maybe we could add a unit test that all BatchProcesses implement the interface correctly, but that might be a separate PR.

Also tipp for the future: We now have a [`CHANGELOG.md`](https://github.com/nils-braun/b2luigi/blob/main/CHANGELOG.md) in the repository. You can add changes and bugfixes into the `## [Unreleased]` section of that file which I will then put under a release section and into the github releases notes when I create a new release. Though if you don't do it, I can also do it myself when I create a new release. Which I will do now, since on Friday I just create a new patch release which now contains this bug.

PR #175, opened 2022-10-31
Parametrize DIRAC group

Bilokin:
Hi @meliache, 

recently, the group policy in gbasf2 was diversified, so the users can now submit the jobs from different DIRAC groups, which have different queue priorities. 
And since there was created a group `belle_systematics`  we would like to add the corresponding functionality to the b2luigi, which is extremely useful for the systematics framework.
To submit a project from a group other than `belle`, one has to do the following:
* Create proxy with the custom group
* Add a custom path to the `gbasf2` command via `--output_path`
* Project download should be done with the same custom DIRAC group
* The project name should be explicitly appended to the temporary  download path, because it is not added by default for a custom group

In this PR I added two optional b2luigi parameters, `gbasf2_proxy_group` and  `gbasf2_project_lpn_path`. 
The latter one is required if the first one is supplied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 * [#166](https://github.com/nils-braun/b2luigi/pull/166): add automatic need-changelog PR labeller as github workflow
+* [#175](https://github.com/nils-braun/b2luigi/pull/175): added ``gbasf2_proxy_group`` and ``gbasf2_project_lpn_path`` parameters to switch between gbasf2 groups.
 
 **Full Changelog**: https://github.com/nils-braun/b2luigi/compare/v0.7.6...main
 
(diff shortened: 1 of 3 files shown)


meliache:
Cool, thanks 🙏. Just looking from my phone this looks good to me. I suggest also documenting this in the class comment for the `Gbasf2Process` at the top of the file, somewhere there is a listing of the important settings for gbasf2 and this will appear on the online sphinx docs. And also add this to the changelog.md under the |`Unpublished` (meaning not to yet in a release) section. The latter I can also do as a maintainer.

Bilokin:
Thanks, I wrote the docs and the entry in the changelog, I hope it is correct. 
There is an issue with the workflow though and I have no clues about it. Can you take a look?

Bilokin:
Hi @meliache, can you take a look at the github workflow issue and the PR in general?

meliache:
Sorry @Bilokin, last week I just had no time to look at this because we were aiming for unblinding of our analysis for PRL, so I didn't look at this properly yet.

meliache:
The flake8 pre-commit hook seems to fail due to an incompatibility between flake8 v3.9, which was set in the pre-commit config, and `importlib_metadata` version 5. I found this by googling as https://github.com/python/importlib_metadata/issues/406. I created a PR which upgrades the pre-commit flake8 version to the latest version in #176. This will hopefully  fix this issue. If not, I might need to further set `importlib_metadata < 5.0.0` in the github action configuration

PR #178, opened 2022-11-10
Fix the wrong tmp path for custom group

Bilokin:
Hi @meliache,

unfortunately, during my last tests of the custom gbasf2 group functionality I missed the fact that the b2luigi produced empty files as output. 
This was because the temporary project name was wrong and, apparently, I tested the functionality on my own workflow, which uses `hadd` and it just creates an empty file when there is no input files. 
This PR should fix the bug for the custom gbasf2 group.

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -669,10 +669,12 @@ class Gbasf2Process(BatchProcess):
             # If the download had been successful and the local files are identical to the list of files on the grid,
             # we move the downloaded dataset to the location specified by ``output_dir_path``.
             tmp_output_dir_path = f"{output_dir_path}.partial"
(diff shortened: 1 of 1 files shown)


Bilokin:
This is related to a weird behaviour in hadd, where the line
```
hadd -f test.root
```
doesn't cause any errors, despite that the second argument is missing.

meliache:
Thanks for the fix and sorry for the delay in reviewing, was quite busy last week and recovered on the weekend by taking time away from my notebook. I removed the "needs a changelog" label because this doesn't change any functionality between releases, just fixes a functionality already described in the changelog for the next releases.

Speaking of that, I bump b2luigi to the next release, which I had been procrastinating for a while. This will make this feature installable via `pip3 install b2luigi` and also make your documentation visible more visible online on readthedocs.

meliache:
This is now released in [v0.8.1](https://github.com/nils-braun/b2luigi/releases/tag/v0.8.1) on github and PyPi.

PR #181, opened 2023-01-12
Fix queries for gbasf2 grid files with multiple extensions

### LINKED ISSUE #180: Hardcoded extension ".root" in _get_gbasf2_dataset_query in b2luigi/batch/processes/gbasf2.py
When using b2luigi with gbasf2 jobs with udst outputs, these will have the ending ".udst.root".  Then the download gbasf2_data_query will be faulty:

`f"sub*/{output_file_stem}_*{output_file_ext}"= sub*/filename.udst_*.root`

instead of `"sub*/filename_*.udst.root"`. This will cause the download to fail.

meliache:
Hotfix that hopefully resolves issue #180, if I understood the behavior of gbasf2 correctly. Currently I don't have time and no access to the grid. @schmitca, could you please check if this works?

@schmitca For installing this branch, you can use
``` bash
pip3 install --upgrade "git+https://github.com/nils-braun/b2luigi#bugfix/gbasf2-correctly-handle-multi-extension-outputs" 
```

Resolves #180

### DIFF
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -583,20 +583,23 @@ class Gbasf2Process(BatchProcess):
         """
         if output_file_name != os.path.basename(output_file_name):
             raise ValueError(
-                f"For grid projects, the output file name must not be a basename, not a path, but is \"{output_file_name}\""
+                f'For grid projects, the output file name must be a basename, not a path, but is "{output_file_name}"'
             )
-        output_file_stem, output_file_ext = os.path.splitext(output_file_name)
-        if output_file_ext != ".root":
+        # split file basename into stem and all extensions (e.g. ".udst.root")
+        output_file_stem, output_file_extensions = output_file_name.split(
+            ".", maxsplit=1
+        )
+        # make sure that last extension is ".root"
+        if os.path.splitext(output_file_extensions)[-1] != ".root":
             raise ValueError(
-                f"Output file name \"{output_file_name}\" does not end with \".root\", "
+                f'Output file name "{output_file_name}" does not end with ".root", '
                 "but gbasf2 batch only supports root outputs"
             )
         output_lpn_dir = f"/belle/user/{self.dirac_user}"
         group_name = get_setting("gbasf2_proxy_group", default="belle")
         if group_name != "belle":
             output_lpn_dir = get_setting("gbasf2_project_lpn_path")
-        dataset_query_string = \
-            f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_ext}"
+        dataset_query_string = f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_extensions}"
         return dataset_query_string
 
     def _local_gb2_dataset_is_complete(self, output_file_name: str, check_temp_dir: bool = False) -> bool:
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -599,7 +599,7 @@ class Gbasf2Process(BatchProcess):
         group_name = get_setting("gbasf2_proxy_group", default="belle")
         if group_name != "belle":
             output_lpn_dir = get_setting("gbasf2_project_lpn_path")
-        dataset_query_string = f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_extensions}"
+        dataset_query_string = '.'.join([f'{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*', output_file_extensions])
         return dataset_query_string
 
     def _local_gb2_dataset_is_complete(self, output_file_name: str, check_temp_dir: bool = False) -> bool:
--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -599,8 +599,9 @@ class Gbasf2Process(BatchProcess):
         group_name = get_setting("gbasf2_proxy_group", default="belle")
         if group_name != "belle":
             output_lpn_dir = get_setting("gbasf2_project_lpn_path")
-        dataset_query_string = '.'.join([f'{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*', output_file_extensions])
-        return dataset_query_string
+        return '.'.join(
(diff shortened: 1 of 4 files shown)


b2luigi/batch/processes/gbasf2.py
```diff
@@ -583,20 +583,23 @@ def _get_gbasf2_dataset_query(self, output_file_name: str) -> str:
...
+                f'Output file name "{output_file_name}" does not end with ".root", '
                 "but gbasf2 batch only supports root outputs"
             )
         output_lpn_dir = f"/belle/user/{self.dirac_user}"
         group_name = get_setting("gbasf2_proxy_group", default="belle")
         if group_name != "belle":
             output_lpn_dir = get_setting("gbasf2_project_lpn_path")
-        dataset_query_string = \
-            f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_ext}"
+        dataset_query_string = f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_extensions}"
```
schmitca, line 602: Since str.split(".")  removes the separation identifer "." this should be changed to 
dataset_query_string = f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*.{output_file_extensions}"

schmitca:
Dear @meliache, I am happy to test it. At the moment, the grid jobs with udst.root output fail, before download can be attempted. This is because running the picked basf2 path with udst.add_skimmed_udst_output(skimDecayMode='skimname') fails with

```
[FATAL] identical from/to parameter value   { module: SwitchDataStore ('' -> 'skim_skimname') function: virtual void Belle2::SwitchDataStoreModule::initialize() }
[INFO] ===Error Summary================================================================
[FATAL] identical from/to parameter value   { module: SwitchDataStore ('' -> 'skim_skimname') }
```

This can be traced back to the fact that basf2.pickle_path.deserialize_path( basf2.pickle_path.serialize_path(mypath)) != mypath. I will try to find/contact an expert for basf2.pickle_path.

b2luigi/batch/processes/gbasf2.py
meliache: Thanks for the suggestion and checking the code, I didn't think about that. Btw, the inverse operation of `split` is `join`, e.g. I can achieve the same with
```python
'.'.join(f'{output_file_stem}_*', output_file_extensions}
```
This is useful when you split into multiple elements and then join multiple elements back together, in this case your solution might be more simple.

meliache:
> Dear @meliache, I am happy to test it. At the moment, the grid jobs with udst.root output fail, before download can be attempted. This is because running the pickled basf2 path with udst.add_skimmed_udst_output(skimDecayMode='skimname') fails

This seems to be a separate issue and I'm afraid if we need to add a workaround for that, that would be quite involved and probably should be a separate pull request, but not sure if I have time for that.
 
> This can be traced back to the fact that `basf2.pickle_path.deserialize_path( basf2.pickle_path.serialize_path(mypath)) != mypath` I will try to find/contact an expert for basf2.pickle_path.

Honestly, good luck with that, I don't expect that to be simple. I think (though don't quote me on that) that Thomas Keck had contributed to the pickle path functions because he needed it for the FEI. Martin Ritter also new about them. But those people left the collaboration alreayd long ago. With those things you are usually more succesful by trying to dig into the code yourself and maybe even contribute.

I just checked the definition of `udst.add_skimmed_udst_output` and it seems that it creates an additional conditional path

``` python
    skim_path = basf2.create_path()
    [...]
    filter_path.add_independent_path(skim_path, "skim_" + skimDecayMode)
    add_udst_output(
        skim_path,
        outputFile,
        saveParticleLists,
        additionalBranches,
        dataDescription=dataDescription,
        mc=mc,
    )
```

I don't understand conditional paths super well, but it's well possible that `serialize_path` has problems with that.
> `basf2.pickle_path.deserialize_path( basf2.pickle_path.serialize_path(mypath)) != mypath`
You could check really what's the difference by eye or looking at the diff of `basf2.print_path`. Sadly I don't think I have the time to dig into thhat.

For testing this pull request I would try just using `udst.add_udst_output`, or also just add dots to a root file name, like `foo.bar.root`. And try figure out the conditional paths seperately

schmitca:
I can confirm that the download is working with udst.add_udst_output and
```

--- a/b2luigi/batch/processes/gbasf2.py
+++ b/b2luigi/batch/processes/gbasf2.py
@@ -599,7 +599,7 @@ class Gbasf2Process(BatchProcess):
         group_name = get_setting("gbasf2_proxy_group", default="belle")
         if group_name != "belle":
             output_lpn_dir = get_setting("gbasf2_project_lpn_path")
-        dataset_query_string = f"{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*{output_file_extensions}"
+        dataset_query_string = '.'.join([f'{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*', output_file_extensions])
         return dataset_query_string
```

It seems that I cannot push my changes, but this can be commited and closed.

meliache:
> I can confirm that the download is working with udst.add_udst_output and

Nice :+1: 

> It seems that I cannot push my changes, but this can be commited and closed.

Thanks for the patch. Actually github has a feature where when the reviewer comments on a line of code, they can add a _code suggestion_ and I could then accept this suggestion. You could do that in the thread on the missing dot bug above. But I could also just do the commit myself, git allows adding other people as the author of a commit (each commit can have a separate commiter and author.

Regarding that, I think I fund your full name and LMU email on b2mms. If I put you as an author of a commit, I'd like to use that and if you allow me to I would also add your full name to the [contributors list](https://b2luigi.readthedocs.io/en/stable/#the-team). I'm asking in case  you don't want that for privacy reasons, but if possible I'd like to credit you properly.

b2luigi/batch/processes/gbasf2.py
schmitca:
```suggestion
        dataset_query_string = '.'.join([f'{output_lpn_dir}/{self.gbasf2_project_name}/sub*/{output_file_stem}_*', output_file_extensions])
```

schmitca:
Thank you @meliache, I added a suggestion.

meliache:
@schmitca thanks for doing that so late. Can I use your full name from B2MMS in the contriburs list in the documentation?

schmitca:
Sure, thank you! @meliache

meliache:
I now published this to PyPi as  release 0.8.2, so if you install b2luigi with `pip install b2luigi --upgrade`, you should have the functionality available immediately.

PR #189, opened 2023-03-29
add ability to pass a custom hash function

MarcelHoh:
I often find myself getting a bit annoyed at just how human unfriendly the hashed parameters are. In many cases, while the set of parameters might be a large dictionary,  I can produce a unique string which allows me to identify the exact parameters passed, or it would be useful to just prepend a little information to a hash to make it easier to identify broad groups of tasks.

This PR adds the ability to use a custom hash function, for hashed parameters. This can be as easy as just joining the values of a list parameter with ```"_".join(x)```. It is up to the user to ensure that no duplicates are created and that names are of a reasonable length. The default behaviour remains unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Unreleased
 
+### Added
+
+- added the ability to pass a custom hashing function to parameters via the `hash_function` keyword argument. [#189](https://github.com/nils-braun/b2luigi/pull/189). The function must take one argument, the value of the parameter. It is up to the user to ensure unique strings are created.
+
 ## [0.9.0] - 2023-03-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Unreleased
 
+### Added
+
+- added the ability to pass a custom hashing function to parameters via the `hash_function` keyword argument. [#189](https://github.com/nils-braun/b2luigi/pull/189). The function must take one argument, the value of the parameter. It is up to the user to ensure unique strings are created.
```
meliache, line 13: Present tense
```suggestion
- Add the ability to pass a custom hashing function to parameters via the `hash_function` keyword argument. [#189](https://github.com/nils-braun/b2luigi/pull/189). The function must take one argument, the value of the parameter. It is up to the user to ensure unique strings are created.
```

PR #190, opened 2023-03-29
switch gb2_ds_get to use the --new flag

MarcelHoh:
Sorry for yet another PR. 

I've been using the --new flag in gb2_ds_get in my local version of b2luigi for the last few months. It is significantly faster than the previous download method and I think it would be good to adopt this officially.

To use the flag gbasf2 v5r6 is required. This has now been out for 5+ months. There may be some holdouts but I think it is probably safe to enable the flag. This is of course up to you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- **gbasf2**: Switched to the `--new` flag in `gb2_ds_get` which downloads files significantly faster than previously. Gbasf2 release v5r6 (November 2022) is required. [#190](https://github.com/nils-braun/b2luigi/pull/190).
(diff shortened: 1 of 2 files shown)


meliache:
Nice to see improvements on that front! Better download handling on gbasf2 side would resolve many issues.

I'm just worried that there will be a new gbasf2 release in the future which might adopt the new download method as a default and deprecate `--new`  parameter, requiring a new PR to remove it again. Though I hope the gbasf2 devs would then just ignore that parameter for backwards compatibility. Are you aware whether it's planned to be enabled by default in the future?

Also, does the new download method still allow for writing the list of files for which the download failed to a file? B2luigi has some code that repeats a failed download using the list of files from that file. The code was mainly introduced for performance reasons, and I hope with better downloads we might deprecate it.

MarcelHoh:
I'm not aware of any definitive plans, like you I would expect that the new flag is not removed (at least not immediately) even if it is no longer necessary. 
I will ask Michel but he is currently on holiday so I will wait until he is back. 

For the failed download list, I haven't encountered any issues with this in the 5 months I've used the `new` flag and b2luigi. But I'll also double check with Michel.

PR #193, opened 2023-03-31
Better basf2-related docs, examples and change how basf2 release hash is calculated for local releases.

GiacomoXT:
Top-level `import ROOT` is very nasty and it must be avoided as much as possible. So, I removed them.

Moreover, better not to invoke the Environment singleton, but to directly use the parmaeter in `basf2.process` for the maximum number of events.

And in case it's not evident, I don't like `basf2.create_path()`. :D

### DIFF
--- a/b2luigi/basf2_helper/tasks.py
+++ b/b2luigi/basf2_helper/tasks.py
@@ -46,21 +46,17 @@ class Basf2PathTask(Basf2Task):
 
         try:
             import basf2
-            import ROOT
         except ImportError:
             raise ImportError("Can not find ROOT or basf2. Can not use the basf2 task.")
 
         if self.num_processes:
             basf2.set_nprocesses(self.num_processes)
 
-        if self.max_event:
-            ROOT.Belle2.Environment.Instance().setNumberEventsOverride(self.max_event)
-
         path = self.create_path()
 
         path.add_module("Progress")
         basf2.print_path(path)
-        basf2.process(path)
+        basf2.process(path=path, max_event=self.max_event if self.max_event else 0)
 
         print(basf2.statistics)
 
--- a/b2luigi/basf2_helper/utils.py
+++ b/b2luigi/basf2_helper/utils.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 
 
 def get_basf2_git_hash():
@@ -21,5 +22,9 @@ def get_basf2_git_hash():
         if hasattr(basf2.version, "get_version"):
             return basf2.version.get_version()
         return basf2.version.version
-    except ImportError:
+    except ImportError as err:
+        warnings.warn(
+            f"No basf2 was found. Setting basf2 git hash to \"not_set\": \n {err}",
+            category=ImportWarning,
+        )
         return "not_set"
--- a/b2luigi/basf2_helper/utils.py
+++ b/b2luigi/basf2_helper/utils.py
@@ -2,17 +2,24 @@ import os
 
 
 def get_basf2_git_hash():
+    """Return name of basf2 release or if local basf2 is used its version hash.
+
+    The version is equivalent to the version returned by ``basf2 --version``.
+
+    Returns ``\"not set\"``, if no basf2 release is set and basf2 cannot be imported.
+    """
     basf2_release = os.getenv("BELLE2_RELEASE")
 
-    if basf2_release == "head" or basf2_release is None:
-
-        try:
-            from basf2.version import get_version
-            basf2_hash = get_version
-        except ImportError:
-            from basf2.version import version
-            basf2_hash = version
-
-        basf2_release = basf2_hash()
-
-    return basf2_release
+    if basf2_release not in ("head", None):
+        return basf2_release
+    # else if BASF2_RELEASE environment variable is not set, get version hash of
+    # local basf2
+    try:
+        import basf2.version
+        # try both get_version method and version attribute, one of which
+        # should work depending on basf2 release
+        if hasattr(basf2.version, "get_version"):
+            return basf2.version.get_version()
+        return basf2.version.version
+    except ImportError:
+        return "not_set"
--- a/b2luigi/basf2_helper/utils.py
+++ b/b2luigi/basf2_helper/utils.py
@@ -1,16 +1,18 @@
 import os
 
-import git
-
 
 def get_basf2_git_hash():
     basf2_release = os.getenv("BELLE2_RELEASE")
 
     if basf2_release == "head" or basf2_release is None:
-        basf2_release_location = os.getenv("BELLE2_LOCAL_DIR")
 
-        if basf2_release_location:
-            return git.Repo(basf2_release_location).head.object.hexsha
-        return "not_set"
+        try:
+            from basf2.version import get_version
+            basf2_hash = get_version
+        except ImportError:
+            from basf2.version import version
+            basf2_hash = version
+
+        basf2_release = basf2_hash()
 
     return basf2_release
--- a/examples/basf2/basf2_chain_example.py
+++ b/examples/basf2/basf2_chain_example.py
@@ -10,7 +10,7 @@ import simulation
 import vertex
 import generators
 import reconstruction
-from ROOT import Belle2
+import mdst
 
 
 class SimulationType(Enum):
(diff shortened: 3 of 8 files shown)


meliache:
Looks good to me :+1: . It's just an example, so if you say you tested it than I'll believe it, I'll just way for the code style checks to finish.

> And in case it's not evident, I don't like basf2.create_path(). :D

`path = basf2.Path()` is definitely more pythonic and I'm in favor of using this, calling the `__init__` via the name of the type to instantiate the object instead of a separate factory method. Just a question out of curiosity, was the possibility to call `basf2.Path` added in some specific basf2 release or was it always there? Around five years ago when I started at Belle II everyone in my group seemed to use `basf2.create_path` and it seemed widespread in examples, so I'm wondering if it was introduced in some point.

examples/basf2/basf2_chain_example.py
```diff
@@ -23,23 +23,22 @@ class SimulationTask(Basf2PathTask):
...
         if self.event_type == SimulationType.y4s:
             # in current main branch and release 5 the Y(4S)decay file is moved, so try old and new locations
             find_file_ignore_error = True
-            dec_file = Belle2.FileSystem.findFile('analysis/examples/tutorials/B2A101-Y4SEventGeneration.dec',
-                                                  find_file_ignore_error)
-            if not dec_file:
-                dec_file = Belle2.FileSystem.findFile('analysis/examples/simulations/B2A101-Y4SEventGeneration.dec')
+            dec_file = basf2.find_file('analysis/examples/tutorials/B2A101-Y4SEventGeneration.dec',
+                                       silent=find_file_ignore_error)
+            if dec_file == '':
```
meliache, line 34: Why explicitly check for an empty string? I assume that's what `basf2.find_file` returns if no file was found? Empty strings also evaluate to False in Python, I'm sure you know. I'm not saying this is wrong, just want to understand the reason.

I'm not sure if there are cases where it returns something other that evaluates to `False` like `None`. Probably not. But maybe we can be paranoid and raise a `ValueError` or `TypeError`  if it returns something other that was covered by the `not` but not by the `== ''`.

b2luigi/basf2_helper/tasks.py
```diff
@@ -46,21 +46,17 @@ def process(self):
 
         try:
             import basf2
-            import ROOT
         except ImportError:
             raise ImportError("Can not find ROOT or basf2. Can not use the basf2 task.")
```
meliache, line 50:
```suggestion
            raise ImportError("Can not find basf2. Can not use the basf2 task.")
```

b2luigi/basf2_helper/tasks.py
```diff
@@ -46,21 +46,17 @@ def process(self):
...
 
-        if self.max_event:
-            ROOT.Belle2.Environment.Instance().setNumberEventsOverride(self.max_event)
-
         path = self.create_path()
 
         path.add_module("Progress")
         basf2.print_path(path)
-        basf2.process(path)
+        basf2.process(path=path, max_event=self.max_event if self.max_event else 0)
```
meliache, line 59: This is a bit nit-picky but I would move this inline if-else into it's separate line. 

```suggestion
        max_event = self.max_event if self.max_event else 0
        basf2.process(path=path, max_event=max_event)
```
Optionally even use a real if-else instead of ternary operator style:
```suggestion
        if self.max_event:
            max_event = self.max_event
        else:
            max_event else 0
        basf2.process(path=path, max_event=max_event)
```
 The reason is that I think code is more readable, yields better diffs and is easier to debug if every piece of logic is on its own line. If too many separate happen in one line it's more difficult to reason about it, though I use that style style in my own scripts sometimes. But few lines of code are not a goal in itself.
 
 From the above two suggestions, choose whichever you like more or refactor that yourself if you have an even better idea.

GiacomoXT:
> `path = basf2.Path()` is definitely more pythonic and I'm in favor of using this, calling the `__init__` via the name of the type to instantiate the object instead of a separate factory method. Just a question out of curiosity, was the possibility to call `basf2.Path` added in some specific basf2 release or was it always there? Around five years ago when I started at Belle II everyone in my group seemed to use `basf2.create_path` and it seemed widespread in examples, so I'm wondering if it was introduced in some point.

`basf2.Path()` is available thanks to https://gitlab.desy.de/belle2/software/basf2/-/blob/main/framework/core/src/Path.cc#L243 and https://gitlab.desy.de/belle2/software/basf2/-/blob/main/framework/scripts/basf2/core.py#L31 and it's there since many years. I think that `basf2.create_path()` was introduced at the beginning when the python-binding was rudimentary (it is in my to-do list for being removed since unnecessary).

examples/basf2/basf2_chain_example.py
GiacomoXT: With `silent=True`, `find_file` returns `''` if nothing is found. With `silent=False`, a `FileNotFoundError` exception is raised.

Why did I use `== ''`? Because I don't know Python. :D

examples/basf2/basf2_chain_example.py
```diff
@@ -56,10 +55,12 @@ def create_path(self):
         path = basf2.create_path()
 
         path.add_module('RootInput', inputFileNames=self.get_input_file_names("simulation_full_output.root"))
-        modularAnalysis.loadGearbox(path)
```
GiacomoXT, line 59: One remark: I am 100% sure that this example was not working, since `add_reconstruction` needs the `Geometry` module for running and `modularAnalysis.loadGearbox` only appends the `Gearbox` module.

(Moreover, we plan to remove `modularAnalysis.loadGearbox`, since there are no reasons for using such module at analysis level.)

b2luigi/basf2_helper/utils.py
```diff
@@ -1,16 +1,18 @@
...
 def get_basf2_git_hash():
     basf2_release = os.getenv("BELLE2_RELEASE")
 
     if basf2_release == "head" or basf2_release is None:
-        basf2_release_location = os.getenv("BELLE2_LOCAL_DIR")
 
-        if basf2_release_location:
-            return git.Repo(basf2_release_location).head.object.hexsha
-        return "not_set"
+        try:
```
GiacomoXT, line 9: Since I was here...

(In this way, `basf2_release` is equivalent to the `version` returned by `basf2 --info` or `basf2 --version`.)

meliache:
By the way, this PR still needs an entry to the `CHANGELOG.md`. I can do that and if you don't mind add you to the contributors list.

> `basf2.create_path()` was introduced at the beginning when the python-binding was rudimentary (it is in my to-do list for being removed since unnecessary).

Sounds sensible to me. In Python it's common to first show a `DeprecationWarning` in one release before deprecating it in the next major release, I would suggest doing something like that.

b2luigi/basf2_helper/utils.py
meliache: I the reasoning, but, in some sense, this is breaking backwards-compatibility from a user-perspective, as it changes the job output target path. That is, once updating, your jobs that were once _complete_ might suddenly not be _complete_ anymore. We're in beta anyway but I would probably change the second digit of the release for this and not just the thirds patch-digit.

I looked into the code and this seems to call the C++ function `getCommitID`:
https://github.com/belle2/basf2/blob/1c972b2c89ef11f38ee2f5ea8eb562dde0637155/framework/io/include/RootIOUtilities.h#L77-L84
> Return git SHA1 hashes taking into account local & central release.

This could have been a separate PR. Fine with me having it here, but I would make this a prominent line in the `Changed` section of the Changelog

examples/basf2/basf2_chain_example.py
meliache: Thanks, I reverted the empty string check to `if not dec_file` in b1ff807c9199d6decd560a89dd91f12bd34a5632 and added your explanation from this comment as a code-comment.

b2luigi/basf2_helper/utils.py
meliache: I was about to merge but found that this change is breaking unittests due to failed import of `basf2`, see https://github.com/nils-braun/b2luigi/actions/runs/4595212940/jobs/8115187454?pr=193.

I think this is because previously, if no basf2 release is set, `get_basf2_git_hash` just returned the string `"not set"` if no release is set. You can argue in favor of raising an error then, but then we would need to fix the unit tests. This might be done by using `unittest.Mock` to mock a fake `basf2.version` package with a `get_version` function or mock the entire `get_basf2_git_hash` function.

- [ ] Fix unittests for `get_basf2_git_hash`, either return `"not set"` in case of failure or fix unittests by using mocking.

meliache: I now reverted the behavior to returning `"not set"` when basf2 can't be imported in 0180a725ed8d2ee6ae7f8b2f5a5f13dbfbe39419, but now additionally print a warning because that's really undesired behavior usually. I also set up unit tests which document the functionality of `get_basf2_git_hash` in 374944c69bf3d5acce702bfba89d107835aee679 andbeaa520f688708e74cf5f0a24726c4804af2807e

GiacomoXT:
@meliache thanks a lot for having taken care of the last remaining issues (I was on vacation during the last two days).

PR #203, opened 2023-05-25
Add gbasf2_setup_path as an argument to several functions requiring it

AlexanderHeidelbach:
Some functions did not propagate the gbasf2_setup_path correctly to all functions in their execution. This causes the default gbasf2 setup to be invoked instead of the custom setting. I tested the behaviour with the currently broken gbasf2 setup script and a custom version and now submission is possible with the custom script..


meliache:
Thanks! LGTM :+1: . Just need to add a changelog and add you to the list of contributers, but I can do that too if you set this PR as editable.

Issue #208, opened 2023-11-02
Gbasf2 job status json is `None` in latest Gbasf2 release


meliache:
Originally reported by @0ctagon in https://github.com/nils-braun/b2luigi/issues/206#issuecomment-1790328890, who  saw the following error message
<details><summary>Traceback</summary>
```
INFO: Worker Worker(salt=8402490694, workers=1, host=cc.kek.jp, username=a, pid=255855) was stopped. Shutting down Keep-Alive thread
Traceback (most recent call last):
  File "b2luigi_gridSubmitDL.py", line 128, in <module>
    main()
  File "b2luigi_gridSubmitDL.py", line 117, in main
    b2luigi.process(
  File "/home/belle2/.local/lib/python3.8/site-packages/b2luigi/cli/process.py", line 113, in process
    runner.run_local(task_list, cli_args, kwargs)
  File "/home/belle2/.local/lib/python3.8/site-packages/b2luigi/cli/runner.py", line 46, in run_local
    run_luigi(task_list, cli_args, kwargs)
  File "/home/belle2/.local/lib/python3.8/site-packages/b2luigi/cli/runner.py", line 62, in run_luigi
    luigi.build(task_list, **kwargs)
  File "/home/belle2/.local/lib/python3.8/site-packages/luigi/interface.py", line 239, in build
    luigi_run_result = _schedule_and_run(tasks, worker_scheduler_factory, override_defaults=env_params)
  File "/home/belle2/.local/lib/python3.8/site-packages/luigi/interface.py", line 173, in _schedule_and_run
    success &= worker.run()
  File "/home/belle2/.local/lib/python3.8/site-packages/luigi/worker.py", line 650, in __exit__
    if task.is_alive():
  File "/home/belle2/.local/lib/python3.8/site-packages/b2luigi/batch/processes/__init__.py", line 135, in is_alive
    job_status = self.get_job_status()
  File "/home/belle2/.local/lib/python3.8/site-packages/b2luigi/batch/processes/gbasf2.py", line 319, in get_job_status
    job_status_dict = get_gbasf2_project_job_status_dict(
  File "/home/belle2/.local/lib/python3.8/site-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
  File "/home/belle2/.local/lib/python3.8/site-packages/retry/api.py", line 90, in retry_decorator
[...]

PR #209, opened 2023-11-03
Fix the jobs status script so it works with the new gbasf2 version

eckerpatrick:
There were some changes in the recent gbasf2 update which caused failures in the job status script, as the CS is not anymore automatically initialized by the `userCreds` decorator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 * **gbasf2:** New setting `gbasf2_setup_path` which can be used to customize the path to the gbasf2 setup file directly (default: `"/cvmfs/belle.kek.jp/grid/gbasf2/pro/tools/setup.sh"`). It is a more flexible replacement for the `gbasf2_install_directory` setting, which will be removed in the future, since we can't predict potential name and path changes of the setup script between gbasf2 releases. @meliache [#162](https://github.com/nils-braun/b2luigi/pull/162)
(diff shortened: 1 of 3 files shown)


eckerpatrick:
this fixes https://github.com/nils-braun/b2luigi/issues/208

meliache:
Thanks, that looks great :+1: ! @0ctagon had figured out the changed import paths but was missing `initializeCS()`.

If you don't mind I'll also add commit 4418938e2b7def3f1f392147ff884c8bdea946c0 for better catching unexpected errors of `gbasf2_job_status.py`. In the b2luigi in the subprocess call of `gbasf2_job_status.py` we only checked for a job status of `3`, which happens when `result["OK"]` is False. However, any other unexpected crash of that script resulted in a non-zero exit code which b2luigi ignored and just continued with a garbage string. Therefore, we should also raise error messages for other non-zero exit codes. Not sure if that would have resulted in a better/earlier error message in this particular case though.

What's finally missing is a Changelog entry.

eckerpatrick:
@meliache sounds good to catch also other failures!
Will you also add the changelog in this context or should I add it?

meliache:
@eckerpatrick Oh you're quick, I just added a Changelog entry in addition to 4418938e2b7def3f1f392147ff884c8bdea946c0. Can you just make a quick check that my changes didn't brick anything? They are minor and shouldn't brake anything, but I have no way to run gbasf2 to check anymore and who knows what I might have overlooked.

(These days I started sometimes merging minor hotfixes/improvements without proper code review using my admin authority and because it'd difficult to find reviewers/testers these day. But people get active here when something is broken so I'm happy to exploit that.)

Can you just do a quick check that

PR #210, opened 2023-11-06
fix the ds list script so it works with the new gbasf2 version

0ctagon:
With the new `gbasf2` version, `userCreds` changed path and `initializeCS()` is not automatic anymore. This PR fix the crash happening when `b2luigi` tries to download the jobs output.
Same solution as in #209 .


meliache:
LGTM :thumbsup:. Just changelog missing, that could be combined with the changelog entry for #209 into a single bulletpoint

